PR #6, opened 2023-04-25
Compatible with 3.9 and small changes

DeepanshuA:
This PR has basically these 3 small changes:
1. Compatible with Python 3.9 as well - conversion to Union syntax
2. Add version as a parameter when doing schedule_new_orchestration
3. Add call_named_activity

### DIFF
--- a/durabletask/client.py
+++ b/durabletask/client.py
@@ -20,6 +20,7 @@ from durabletask import task
 TInput = TypeVar('TInput')
 TOutput = TypeVar('TOutput')
 
+
 class OrchestrationStatus(Enum):
     """The status of an orchestration instance."""
     RUNNING = pb.ORCHESTRATION_STATUS_RUNNING


durabletask/client.py
```diff
@@ -92,38 +92,46 @@ def new_orchestration_state(instance_id: str, res: pb.GetInstanceResponse) -> Or
...
         self._stub = stubs.TaskHubSidecarServiceStub(channel)
         self._logger = shared.get_logger(log_handler, log_formatter)
 
-    def schedule_new_orchestration(self, orchestrator: task.Orchestrator[TInput, TOutput], *,
-                                   input: TInput | None = None,
-                                   instance_id: str | None = None,
-                                   start_at: datetime | None = None) -> str:
+    def schedule_new_orchestration(self, orchestrator: Union[task.Orchestrator[TInput, TOutput], str], *,
+                                   input: Union[TInput, None] = None,
+                                   version: Union[str, None] = None,
```
cgillum, line 104: Let's not include version in this API yet. We shouldn't need it.

durabletask/client.py
```diff
@@ -92,38 +92,46 @@ def new_orchestration_state(instance_id: str, res: pb.GetInstanceResponse) -> Or
...
+                                   instance_id: Union[str, None] = None,
+                                   start_at: Union[datetime, None] = None) -> str:
+
+        if isinstance(orchestrator, str):
+            name = orchestrator
+        else:
+            name = task.get_name(orchestrator)
```
cgillum, line 111: Let's make this a one-liner:
```python
name = orchestrator if isinstance(orchestrator, str) else task.get_name(orchestrator)
```

cgillum (commented):
Looks great - thanks for this PR! Just a couple small suggestions.

durabletask/client.py
DeepanshuA: It fails here in this case: https://github.com/microsoft/durabletask-go/blob/91f53a8f8b7fa24a3cb348fb8290594f6ed1c641/backend/executor.go#L286

When testing with dapr, which is currently using v0.1.4 (https://github.com/dapr/dapr/blob/dbeea96710d38d08dcc5aaf6807914a7d697bed4/go.mod#L35), I have following code piece:
```
ctx, span := helpers.StartNewCreateOrchestrationSpan(ctx, req.Name, req.Version.Value, instanceID)
```

cgillum: I see. In that case, let’s fix it by sending an empty string as the version, but let’s not expose the version to our Python APIs.

durabletask/worker.py
```diff
@@ -336,9 +336,13 @@ def create_timer(self, fire_at: datetime | timedelta) -> task.Task:
         return timer_task
 
     def call_activity(self, activity: task.Activity[TInput, TOutput], *,
-                      input: TInput | None = None) -> task.Task[TOutput]:
-        id = self.next_sequence_number()
+                      input: Union[TInput, None] = None) -> task.Task[TOutput]:
         name = task.get_name(activity)
+        return self.call_named_activity(name, activity, input=input)
+
+    def call_named_activity(self, name: str, activity: task.Activity[TInput, TOutput], *,
```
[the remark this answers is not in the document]
DeepanshuA: Done

durabletask/client.py
DeepanshuA: Done

cgillum (approved):

PR #10, opened 2023-05-02
Compatible with 3.7 onwards

DeepanshuA:
This PR aims to make durabletask compatible with 3.7 onwards.


durabletask/worker.py
```diff
@@ -228,8 +228,8 @@ def __init__(self, instance_id: str):
         self._is_replaying = True
         self._is_complete = False
         self._result = None
-        self._pending_actions = dict[int, pb.OrchestratorAction]()
-        self._pending_tasks = dict[int, task.CompletableTask]()
+        self._pending_actions = {}
+        self._pending_tasks = {}
```
cgillum, line 232: Let’s add type hints to these variable declarations.

setup.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import setuptools
```
cgillum, line 6: Why do we need setup.py to support 3.7?

durabletask/worker.py
DeepanshuA: Yeah, added type hints whereever required.

setup.py
DeepanshuA: I have been publishing using setup.py on Pypi, but as discussed I will use pyproject.toml, and will create a separate PR for it. So, removed it.

pyproject.toml
```diff
@@ -9,7 +9,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "durabletask"
-version = "0.1.0"
+version = "0.1.0-alpha.3"
```
DeepanshuA, line 12: Reviewers: Right now, publishing only alpha versions.

.github/workflows/pr-validation.yml
```diff
@@ -27,11 +27,11 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 pytest
+        pip install setuptools wheel tox pytest
         pip install -r requirements.txt
     - name: Lint with flake8
       run: |
-        flake8 . --count --show-source --statistics --exit-zero
+        tox -e flake8
```
cgillum, line 34: Why do we need to replace the `flake8` command with a `tox` command?

cgillum (commented):
Looks good - just one question about tox.

.github/workflows/pr-validation.yml
DeepanshuA: Ohh, I had updated the PR but forgot to comment here. 
As discussed, have updated with an if condition to skip linting check for 3.7 ver. 
And, build is green now.
Please check.

cgillum (approved):
Looks good - thanks!

PR #11, opened 2023-06-16
Retry policies implementation

DeepanshuA:
This PR introduces retry policies.

Retry Policies can have following attributes:
 1. First Retry Interval
 2. Max Number Of Attempts
 3. Backoff Coefficient
 4. Max Retry Interval
 5. Retry Timeout

Once this PR is merged, would need to incorporate Docs also for Retry Policy and attach the Design approach taken for it.

### DIFF
--- a/durabletask/task.py
+++ b/durabletask/task.py
@@ -4,6 +4,7 @@
 # See https://peps.python.org/pep-0563/
 from __future__ import annotations
 
+import math
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
 from typing import (Any, Callable, Generator, Generic, List, Optional, TypeVar,
@@ -266,29 +267,6 @@ class CompositeTask(Task[T]):
     def on_child_completed(self, task: Task[T]):
         pass
 
-
-class CompletableTask(Task[T]):
-
-    def __init__(self):
-        super().__init__()
-
-    def complete(self, result: T):
-        if self._is_complete:
-            raise ValueError('The task has already completed.')
-        self._result = result
-        self._is_complete = True
-        if self._parent is not None:
-            self._parent.on_child_completed(self)
-
-    def fail(self, message: str, details: pb.TaskFailureDetails):
-        if self._is_complete:
-            raise ValueError('The task has already completed.')
-        self._exception = TaskFailedError(message, details)
-        self._is_complete = True
-        if self._parent is not None:
-            self._parent.on_child_completed(self)
-
-
 class WhenAllTask(CompositeTask[List[T]]):
     """A task that completes when all of its child tasks complete."""
 
@@ -316,6 +294,88 @@ class WhenAllTask(CompositeTask[List[T]]):
 
     def get_completed_tasks(self) -> int:
         return self._completed_tasks
+
+
+class CompletableTask(Task[T]):
+
+    def __init__(self):
+        super().__init__()
+        self._retryable_parent = None
+
+    def complete(self, result: T):
+        if self._is_complete:
+            raise ValueError('The task has already completed.')
+        self._result = result
+        self._is_complete = True
+        if self._parent is not None:
+            self._parent.on_child_completed(self)
+
+    def fail(self, message: str, details: pb.TaskFailureDetails):
+        if self._is_complete:
+            raise ValueError('The task has already completed.')
+        self._exception = TaskFailedError(message, details)
+        self._is_complete = True
+        if self._parent is not None:
+            self._parent.on_child_completed(self)
+
+
+class RetryableTask(CompletableTask[T]):
+
+    def __init__(self, retry_policy: RetryPolicy, action: pb.OrchestratorAction) -> None:
+        super().__init__()
+        self._action = action
+        self._retry_policy = retry_policy
+        self._retry_count = 0
+        self.first_attempt: datetime
+
+    def try_completion(self) -> bool:
+        if self._retry_count >= self._retry_policy.max_number_of_attempts - 1:
+            return True
+        else:
+            return False
+    
+    def increment_retry_count(self) -> None:
+        self._retry_count += 1
+    
+    def compute_next_delay_in_seconds(self) -> timedelta:
+        next_delay = timedelta.min
+        retry_expiration: datetime = datetime.max
+        if self._retry_policy.retry_timeout is not None and self._retry_policy.retry_timeout != datetime.max:
+            retry_expiration = self.first_attempt + self._retry_policy.retry_timeout
+        
+        if self._retry_policy.backoff_coefficient is None:
+            backoff_coefficient = 1.0
+        else:
+            backoff_coefficient = self._retry_policy.backoff_coefficient
+
+        if datetime.utcnow() < retry_expiration:
+            next_delay_f = math.pow(backoff_coefficient, self._retry_count) * self._retry_policy.first_retry_interval.total_seconds()
+
+            if self._retry_policy.max_retry_interval is not None:
+                next_delay_f = min(next_delay_f, self._retry_policy.max_retry_interval.total_seconds())
+                return timedelta(seconds=next_delay_f)
+
+        return next_delay
+
+
+class TimerTask(CompletableTask[T]):
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def set_retryable_parent(self, retryable_task: RetryableTask, retryable_type: str):
+        self._retryable_parent = retryable_task
+        self._retryable_type = retryable_type
+
+    def complete(self, result: T):
+        if self._is_complete:
+            raise ValueError('The task has already completed.')
+        self._result = result
+        self._is_complete = True
+        if self._retryable_parent is not None:
+            self._retryable_parent.try_completion()
+        elif self._parent is not None:
+            self._parent.on_child_completed(self)
 
 
 class WhenAnyTask(CompositeTask[Task]):
--- a/durabletask/task.py
+++ b/durabletask/task.py
@@ -301,6 +301,7 @@ class CompletableTask(Task[T]):
     def __init__(self):
         super().__init__()
         self._retryable_parent = None
+        self._retryable_type = None
 
     def complete(self, result: T):
         if self._is_complete:
--- a/durabletask/task.py
+++ b/durabletask/task.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 import math
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
-from enum import Enum
 from typing import (Any, Callable, Generator, Generic, List, Optional, TypeVar,
                     Union)
 
@@ -210,15 +209,6 @@ class OrchestrationStateError(Exception):
     pass
 
 
-class TaskType(Enum):
-    """A enumeration that represents the kind of possible task.
-    """
-    # Activity task
-    ACTIVITY = 0
-    # Sub-orchestrator task
-    SUB_ORCHESTRATOR = 1
-
-
 class Task(ABC, Generic[T]):
     """Abstract base class for asynchronous tasks in a durable orchestration."""
     _result: T
@@ -332,13 +322,13 @@ class CompletableTask(Task[T]):
 class RetryableTask(CompletableTask[T]):
 
     def __init__(self, retry_policy: RetryPolicy, action: pb.OrchestratorAction,
-                 task_type: TaskType) -> None:
+                 is_sub_orch: bool) -> None:
         super().__init__()
         self._action = action
         self._retry_policy = retry_policy
         self._retry_count = 0
         self._first_attempt: datetime
-        self._task_type = task_type
+        self._is_sub_orch = is_sub_orch
 
     def try_completion(self) -> bool:
         if self._retry_count >= self._retry_policy.max_number_of_attempts - 1:
--- a/durabletask/task.py
+++ b/durabletask/task.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import math
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
+from enum import Enum
 from typing import (Any, Callable, Generator, Generic, List, Optional, TypeVar,
                     Union)
 
@@ -209,6 +210,15 @@ class OrchestrationStateError(Exception):
     pass
 
 
+class TaskType(Enum):
+    """A enumeration that represents the kind of possible task.
+    """
+    # Activity task
+    ACTIVITY = 0
+    # Sub-orchestrator task
+    SUB_ORCHESTRATOR = 1
+
+
 class Task(ABC, Generic[T]):
     """Abstract base class for asynchronous tasks in a durable orchestration."""
     _result: T
@@ -301,7 +311,6 @@ class CompletableTask(Task[T]):
     def __init__(self):
         super().__init__()
         self._retryable_parent = None
-        self._retryable_type = None
 
     def complete(self, result: T):
         if self._is_complete:
@@ -322,12 +331,14 @@ class CompletableTask(Task[T]):
 
 class RetryableTask(CompletableTask[T]):
 
-    def __init__(self, retry_policy: RetryPolicy, action: pb.OrchestratorAction) -> None:
+    def __init__(self, retry_policy: RetryPolicy, action: pb.OrchestratorAction,
+                 task_type: TaskType) -> None:
         super().__init__()
         self._action = action
         self._retry_policy = retry_policy
         self._retry_count = 0
         self._first_attempt: datetime
+        self._task_type = task_type
 
     def try_completion(self) -> bool:
         if self._retry_count >= self._retry_policy.max_number_of_attempts - 1:
@@ -367,9 +378,8 @@ class TimerTask(CompletableTask[T]):
     def __init__(self) -> None:
         super().__init__()
 
-    def set_retryable_parent(self, retryable_task: RetryableTask, retryable_type: str):
+    def set_retryable_parent(self, retryable_task: RetryableTask):
         self._retryable_parent = retryable_task
-        self._retryable_type = retryable_type
 
     def complete(self, result: T):
         if self._is_complete:
--- a/durabletask/task.py
+++ b/durabletask/task.py
@@ -327,7 +327,7 @@ class RetryableTask(CompletableTask[T]):
         self._action = action
         self._retry_policy = retry_policy
         self._retry_count = 0
-        self.first_attempt: datetime
+        self._first_attempt: datetime
 
     def try_completion(self) -> bool:
         if self._retry_count >= self._retry_policy.max_number_of_attempts - 1:
@@ -335,6 +335,9 @@ class RetryableTask(CompletableTask[T]):
         else:
             return False
     
+    def set_first_attempt(self, first_attempt: datetime) -> None:
+        self._first_attempt = first_attempt
+
     def increment_retry_count(self) -> None:
         self._retry_count += 1
     
@@ -342,7 +345,7 @@ class RetryableTask(CompletableTask[T]):
         next_delay = timedelta.min
         retry_expiration: datetime = datetime.max
         if self._retry_policy.retry_timeout is not None and self._retry_policy.retry_timeout != datetime.max:
-            retry_expiration = self.first_attempt + self._retry_policy.retry_timeout
+            retry_expiration = self._first_attempt + self._retry_policy.retry_timeout
         
         if self._retry_policy.backoff_coefficient is None:
             backoff_coefficient = 1.0
--- a/durabletask/task.py
+++ b/durabletask/task.py
@@ -342,7 +342,7 @@ class RetryableTask(CompletableTask[T]):
     def increment_retry_count(self) -> None:
         self._retry_count += 1
     
-    def compute_next_delay_in_seconds(self) -> timedelta:
+    def compute_next_delay(self) -> timedelta:
         next_delay = timedelta.min
         retry_expiration: datetime = datetime.max
         if self._retry_policy.retry_timeout is not None and self._retry_policy.retry_timeout != datetime.max:
--- a/durabletask/task.py
+++ b/durabletask/task.py
@@ -320,33 +320,38 @@ class CompletableTask(Task[T]):
 
 
 class RetryableTask(CompletableTask[T]):
+    """
+    A task that can be retried according to a retry policy.
+    _retry_count variable is initialized to 0 and incremented each time the task is retried.
+    _retry_count is used to determine if the task has exhausted its retries. It will be compared to the max_number_of_attempts
+    in the retry policy and if it is greater than or equal to max_number_of_attempts - 1, the task will not be retried.
+    _retry_count is always 1 less than the total number of overall attempts done.
+    """
 
     def __init__(self, retry_policy: RetryPolicy, action: pb.OrchestratorAction,
-                 is_sub_orch: bool) -> None:
+                 start_time:datetime, is_sub_orch: bool) -> None:
         super().__init__()
         self._action = action
         self._retry_policy = retry_policy
-        self._retry_count = 0
-        self._first_attempt: datetime
+        self._attempt_count = 1
+        self._start_time = start_time
         self._is_sub_orch = is_sub_orch
 
-    def try_completion(self) -> bool:
-        if self._retry_count >= self._retry_policy.max_number_of_attempts - 1:
+    def retry_exhausted(self) -> bool:
+        if self._attempt_count >= self._retry_policy.max_number_of_attempts:
             return True
         else:
             return False
-    
-    def set_first_attempt(self, first_attempt: datetime) -> None:
-        self._first_attempt = first_attempt
 
-    def increment_retry_count(self) -> None:
-        self._retry_count += 1
+    def increment_attempt_count(self) -> None:
+        self._attempt_count += 1
     
-    def compute_next_delay(self) -> timedelta:
-        next_delay = timedelta.min
+    def compute_next_delay(self) -> Union[timedelta, None]:
+        if self._attempt_count >= self._retry_policy.max_number_of_attempts:
+            return None
         retry_expiration: datetime = datetime.max
         if self._retry_policy.retry_timeout is not None and self._retry_policy.retry_timeout != datetime.max:
-            retry_expiration = self._first_attempt + self._retry_policy.retry_timeout
+            retry_expiration = self._start_time + self._retry_policy.retry_timeout
         
         if self._retry_policy.backoff_coefficient is None:
             backoff_coefficient = 1.0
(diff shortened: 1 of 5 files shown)


durabletask/worker.py
```diff
@@ -353,44 +353,92 @@ def is_replaying(self) -> bool:
     def current_utc_datetime(self, value: datetime):
         self._current_utc_datetime = value
 
-    def create_timer(self, fire_at: Union[datetime, timedelta]) -> task.Task:
+    def create_timer(self, fire_at: Union[datetime, timedelta],
+                     retryable_task: Optional[task.RetryableTask] = None,
+                     retryable_type: Optional[str] = None) -> task.Task:
```
cgillum, line 358: Interesting, so in Python, we're allowed to override an abstract method with a concrete method that defines additional (optional) parameters?

durabletask/worker.py
```diff
@@ -353,44 +353,92 @@ def is_replaying(self) -> bool:
...
-        self._pending_actions[id] = action
-
-        activity_task = task.CompletableTask[TOutput]()
-        self._pending_tasks[id] = activity_task
-        return activity_task
+        
+        self.call_activity_util(id, activity, input, retry_policy)
+        return self._pending_tasks.get(id, task.CompletableTask())
+
+    def call_activity_util(self, id: Optional[int],
```
cgillum, line 379: Let's call this `call_activity_helper`. "Util" seems a little unnatural in this context.

durabletask/worker.py
```diff
@@ -353,44 +353,92 @@ def is_replaying(self) -> bool:
...
+                           retry_policy: Optional[task.RetryPolicy] = None,
+                           act_task: Optional[task.CompletableTask[TOutput]] = None):
+            if id is None:
+                id = self.next_sequence_number()
+            name = activity if isinstance(activity, str) else task.get_name(activity)
+                    
+            if act_task is None:
+                encoded_input = shared.to_json(input) if input else None
+            else:
+                encoded_input = input.__str__()
```
cgillum, line 391: Why do we do `__str__()` here? Can you add a comment explaining this?

durabletask/worker.py
```diff
@@ -353,44 +353,92 @@ def is_replaying(self) -> bool:
...
+            self._pending_actions[id] = action
+            if act_task is None:
+                if retry_policy is None:
+                    activity_task = task.CompletableTask[TOutput]()
+                    self._pending_tasks[id] = activity_task
+                else:
+                    activity_task = task.RetryableTask[TOutput](retry_policy=retry_policy, action=action)
+                    self._pending_tasks[id] = activity_task
+            else:
+                self._pending_tasks[id] = act_task
```
cgillum, line 402: It seems like we can reduce some of this logic a bit.
```python
if act_task is None:
    if retry_policy is None:
        act_task = task.CompletableTask[TOutput]()
    else:
        act_task = task.RetryableTask[TOutput](retry_policy=retry_policy, action=action)
self._pending_tasks[id] = act_task
```

durabletask/worker.py
```diff
@@ -353,44 +353,92 @@ def is_replaying(self) -> bool:
...
+                              instance_id: Optional[str] = None,
+                              retry_policy: Optional[task.RetryPolicy] = None) -> task.Task[TOutput]:
         id = self.next_sequence_number()
-        name = task.get_name(orchestrator)
+        orchestrator_name = task.get_name(orchestrator)
+        self.call_sub_orchestrator_util(id, orchestrator_name, input=input, instance_id=instance_id,
+                                        retry_policy=retry_policy)
+        return self._pending_tasks.get(id, task.CompletableTask())
+    
+    def call_sub_orchestrator_util(self, id: Optional[int],
```
cgillum, line 414: Prefer `call_sub_orchestrator_helper`.

durabletask/worker.py
```diff
@@ -353,44 +353,92 @@ def is_replaying(self) -> bool:
...
+        
         if instance_id is None:
             # Create a deteministic instance ID based on the parent instance ID
             instance_id = f"{self.instance_id}:{id:04x}"
-        encoded_input = shared.to_json(input) if input else None
-        action = ph.new_create_sub_orchestration_action(id, name, instance_id, encoded_input)
+        if orch_task is None:
+            encoded_input = shared.to_json(input) if input else None
+        else:
+            encoded_input = input.__str__()
```
cgillum, line 429: Same comment about documenting why we use `__str__()` here.

durabletask/worker.py
```diff
@@ -353,44 +353,92 @@ def is_replaying(self) -> bool:
...
-        return sub_orch_task
+        if orch_task is None:
+            if retry_policy is None:
+                sub_orch_task = task.CompletableTask[TOutput]()
+                self._pending_tasks[id] = sub_orch_task
+            else:
+                sub_orch_task = task.RetryableTask[TOutput](retry_policy=retry_policy, action=action)
+                self._pending_tasks[id] = sub_orch_task
+        else:
+            self._pending_tasks[id] = orch_task
```
cgillum, line 441: Same comment about simplifying logic.

durabletask/worker.py
```diff
@@ -517,12 +564,32 @@ def process_event(self, ctx: _RuntimeOrchestrationContext, event: pb.HistoryEven
                             f"{ctx.instance_id}: Ignoring unexpected timerFired event with ID = {timer_id}.")
                     return
                 timer_task.complete(None)
-                ctx.resume()
+                if timer_task._retryable_parent is not None and timer_task._retryable_type is not None:
+                    activity_action = timer_task._retryable_parent._action
+                    
+                    if timer_task._retryable_type == "activity":
```
cgillum, line 570: I'm wondering if there's a way to simplify the design so that we don't need the `_retryable_type` field. It feels a bit fragile. I don't have any specific suggestions right now, but I'd like us to consider it.

durabletask/worker.py
```diff
@@ -517,12 +564,32 @@ def process_event(self, ctx: _RuntimeOrchestrationContext, event: pb.HistoryEven
...
+                else:
+                    ctx.resume()
             elif event.HasField("taskScheduled"):
                 # This history event confirms that the activity execution was successfully scheduled.
                 # Remove the taskScheduled event from the pending action list so we don't schedule it again.
                 task_id = event.eventId
                 action = ctx._pending_actions.pop(task_id, None)
+                activity_task = ctx._pending_tasks.get(task_id, None)
+                if isinstance(activity_task, task.RetryableTask):
+                    activity_task.set_first_attempt(ctx.current_utc_datetime)
```
cgillum, line 592: Shouldn't we be setting this value inside the `call_activity` method, when we initially create the task?

durabletask/worker.py
```diff
@@ -602,17 +690,37 @@ def process_event(self, ctx: _RuntimeOrchestrationContext, event: pb.HistoryEven
                         self._logger.warning(
                             f"{ctx.instance_id}: Ignoring unexpected subOrchestrationInstanceFailed event with ID = {task_id}.")
                     return
-                sub_orch_task.fail(
-                    f"Sub-orchestration task #{task_id} failed: {failedEvent.failureDetails.errorMessage}",
-                    failedEvent.failureDetails)
-                ctx.resume()
+                if isinstance(sub_orch_task, task.RetryableTask):
```
cgillum, line 693: Looks like there's a lot of duplicate logic here. I wonder if there's a way we can share more of the logic so that fixes to retry behavior automatically apply to both activities and sub-orchestrations.

tests/test_orchestration_e2e.py
```diff
@@ -268,3 +268,39 @@ def orchestrator(ctx: task.OrchestrationContext, input: int):
...
+        state = task_hub_client.wait_for_orchestration_completion(id, timeout=30)
+        assert state is not None
+        assert state.runtime_status == client.OrchestrationStatus.FAILED
+        assert state.failure_details is not None
+        assert state.failure_details.error_type == "RuntimeError"
+        assert state.failure_details.message == "Kah-BOOOOM!!!"
+        assert state.failure_details.stack_trace is not None
+        assert state.failure_details.stack_trace.startswith("RuntimeError: Kah-BOOOOM!!!\n")
+
+        # TODO: Verify that the throw_activity was called 9 times and the child_orchestrator was called 3 times
```
cgillum, line 306: Do you plan on completing this test implementation?

tests/test_orchestration_executor.py
```diff
@@ -228,6 +228,96 @@ def orchestrator(ctx: task.OrchestrationContext, orchestrator_input):
...
+    new_events = [
+        helpers.new_orchestrator_started_event(current_timestamp),
+        helpers.new_task_failed_event(5, Exception("Kah-BOOOOM!!!"))]
+    executor = worker._OrchestrationExecutor(registry, TEST_LOGGER)
+    actions = executor.execute(TEST_INSTANCE_ID, old_events, new_events)
+    assert len(actions) == 2
+    assert actions[0].HasField("createTimer")
+    assert actions[0].id == 2
+
+    # TODO: Keep going, and confirm the behavior of max_retry_interval and retry_timeout
```
cgillum, line 318: Do you plan on completing this test implementation?

durabletask/worker.py
DeepanshuA: Yeah, it seems that is allowed in python: https://stackoverflow.com/a/50009140/18848251. But, I will change the interface signature as well, as current state looks like a bad practice. The implementation didn't complain, so I also missed it.

cgillum: But does this extra parameter even make sense to the developer? It's mostly internal, only, right? I don't have a problem with what you're doing here, but I think we probably want to not include `retryable_type` in the interface. Another option would be to just create a new method, like `create_timer_internal` which is called by `create_timer` and by your other code path.

durabletask/worker.py
DeepanshuA: Done, merged both helper methods and calling `call_activity_function_helper`.

durabletask/worker.py
DeepanshuA: Added comment:

```
# Here, we don't need to convert the input to JSON because it is already converted.
# We just need to take string representation of it.
```

durabletask/worker.py
DeepanshuA: Done

durabletask/worker.py
DeepanshuA: Merged both helper methods and calling call_activity_function_helper.

durabletask/worker.py
DeepanshuA: Have tried to modify the logic, to look more standarized. Using an enum and setting it as a part of RetryableTask.

durabletask/worker.py
DeepanshuA: Yeah, that sounds better - have changed.

durabletask/worker.py
DeepanshuA: Modified to use a boolean, instead of enum.

tests/test_orchestration_e2e.py
DeepanshuA: Completed

tests/test_orchestration_executor.py
DeepanshuA: Completed

durabletask/task.py
```diff
@@ -313,6 +296,92 @@ def get_completed_tasks(self) -> int:
...
+        else:
+            return False
+    
+    def set_first_attempt(self, first_attempt: datetime) -> None:
+        self._first_attempt = first_attempt
+
+    def increment_retry_count(self) -> None:
+        self._retry_count += 1
+    
+    def compute_next_delay_in_seconds(self) -> timedelta:
```
cgillum, line 345: The function name ends with `_in_seconds` but the function returns a `timedelta`, which is inconsistent. Consider removing `_in_seconds` from the name.

cgillum (commented):
One thing I'm concerned about with this PR is whether it correctly handles cases where retry policies are mixed with `when_all` and `when_any` use cases. Can we add some test cases that cover these scenarios?

durabletask/task.py
DeepanshuA: Done

durabletask/worker.py
DeepanshuA: `input` here is of type `TInput@call_activity_function_helper`.
And, below `new_schedule_task_action` expects `encoded_input` to be of type `str`.
I have removed `__str__()` but rather typecasted to `str` now.

DeepanshuA:
> One thing I'm concerned about with this PR is whether it correctly handles cases where retry policies are mixed with `when_all` and `when_any` use cases. Can we add some test cases that cover these scenarios?

now, I have added tests for when_any and when_all with retryable task.

durabletask/task.py
```diff
@@ -313,6 +296,92 @@ def get_completed_tasks(self) -> int:
...
+
+class RetryableTask(CompletableTask[T]):
+
+    def __init__(self, retry_policy: RetryPolicy, action: pb.OrchestratorAction,
+                 is_sub_orch: bool) -> None:
+        super().__init__()
+        self._action = action
+        self._retry_policy = retry_policy
+        self._retry_count = 0
+        self._first_attempt: datetime
```
cgillum, line 330: Let's initialize the `_first_attempt` variable so that we don't have any uninitialized fields.

durabletask/task.py
```diff
@@ -313,6 +296,92 @@ def get_completed_tasks(self) -> int:
...
+        self._is_sub_orch = is_sub_orch
+
+    def try_completion(self) -> bool:
+        if self._retry_count >= self._retry_policy.max_number_of_attempts - 1:
```
cgillum, line 334: Instead of tracking the number of retries, would it be more intuitive to track the number of attempts since we're comparing to `max_number_of_attempts`?

durabletask/task.py
```diff
@@ -313,6 +296,92 @@ def get_completed_tasks(self) -> int:
...
+            return False
+    
+    def set_first_attempt(self, first_attempt: datetime) -> None:
+        self._first_attempt = first_attempt
+
+    def increment_retry_count(self) -> None:
+        self._retry_count += 1
+    
+    def compute_next_delay(self) -> timedelta:
+        next_delay = timedelta.min
```
cgillum, line 346: Here you're assigning `next_delay` to `timedelta.min` but aren't referencing it nor are you reassigning it. Seems like this can be removed.

durabletask/task.py
```diff
@@ -313,6 +296,92 @@ def get_completed_tasks(self) -> int:
...
+            backoff_coefficient = self._retry_policy.backoff_coefficient
+
+        if datetime.utcnow() < retry_expiration:
+            next_delay_f = math.pow(backoff_coefficient, self._retry_count) * self._retry_policy.first_retry_interval.total_seconds()
+
+            if self._retry_policy.max_retry_interval is not None:
+                next_delay_f = min(next_delay_f, self._retry_policy.max_retry_interval.total_seconds())
+                return timedelta(seconds=next_delay_f)
+
+        return next_delay
```
cgillum, line 363: It looks like you're returning `timedelta.min` if the retry expiration has passed. I wonder if it would be more intuitive to return `None` in that case as a more clear signal that we cannot delay any further.

durabletask/task.py
```diff
@@ -313,6 +296,92 @@ def get_completed_tasks(self) -> int:
...
+
+class TimerTask(CompletableTask[T]):
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def set_retryable_parent(self, retryable_task: RetryableTask):
+        self._retryable_parent = retryable_task
+
+    def complete(self, result: T):
```
cgillum, line 374: Timer tasks shouldn't be returning any results, so I don't see a need to have this `result` parameter.

durabletask/worker.py
```diff
@@ -602,17 +673,37 @@ def process_event(self, ctx: _RuntimeOrchestrationContext, event: pb.HistoryEven
...
+                                ctx.resume()
+                            else:
+                                ctx.create_timer_internal(next_delay, sub_orch_task)
+                elif isinstance(sub_orch_task, task.CompletableTask):
+                    sub_orch_task.fail(
+                        f"Sub-orchestration task #{task_id} failed: {failedEvent.failureDetails.errorMessage}",
+                        failedEvent.failureDetails)
+                    ctx.resume()
+                else:
+                    raise Exception("Unexpected sub orchestrator type")
```
cgillum, line 699: I think `TypeError` might be more appropriate here. Also, the error message is a bit confusing. I think it should be something like "Unexpected task type".

durabletask/task.py
```diff
@@ -313,6 +296,92 @@ def get_completed_tasks(self) -> int:
...
+    def __init__(self, retry_policy: RetryPolicy, action: pb.OrchestratorAction,
+                 is_sub_orch: bool) -> None:
+        super().__init__()
+        self._action = action
+        self._retry_policy = retry_policy
+        self._retry_count = 0
+        self._first_attempt: datetime
+        self._is_sub_orch = is_sub_orch
+
+    def try_completion(self) -> bool:
```
cgillum, line 333: The name `try_completion` is confusing because it sounds like you're attempting to complete the task. Since you're actually just checking to see if we've exceeded our retry limit, should we rename this to something like `can_retry`?

durabletask/task.py
```diff
@@ -313,6 +296,92 @@ def get_completed_tasks(self) -> int:
...
+        if self._retry_count >= self._retry_policy.max_number_of_attempts - 1:
+            return True
```
cgillum, line 335: Do we also need to consider whether we've reached the retry expiration as part of this method? That might make this method more complete, and then the calling code in worker.py won't have to separately check the retry expiration.

durabletask/worker.py
```diff
@@ -602,17 +673,37 @@ def process_event(self, ctx: _RuntimeOrchestrationContext, event: pb.HistoryEven
...
+                                    f"Sub-orchestration task #{task_id} failed: {failedEvent.failureDetails.errorMessage}",
+                                    failedEvent.failureDetails)
+                                ctx.resume()
+                            else:
+                                ctx.create_timer_internal(next_delay, sub_orch_task)
+                elif isinstance(sub_orch_task, task.CompletableTask):
+                    sub_orch_task.fail(
+                        f"Sub-orchestration task #{task_id} failed: {failedEvent.failureDetails.errorMessage}",
+                        failedEvent.failureDetails)
+                    ctx.resume()
```
cgillum, line 697: It seems we've repeated this `sub_orch_task.fail(...)` and `ctx.resume()` code three times in this block. Can we refactor this at all to reduce the duplication? My earlier suggestion might reduce at least one of these.

tests/test_orchestration_e2e.py
```diff
@@ -268,3 +268,132 @@ def orchestrator(ctx: task.OrchestrationContext, input: int):
...
+    # It does this by creating an orchestration that calls a sub-orchestrator,
+    # which in turn calls an activity that always fails. 
+    # In the first setup, this setup is done without any retry policies, and
+    # the orchestration should fail. Here, sub-orchestrator is called twice,
+    # and the activity is called once.
+    # In the second setup, the retry policies are added, and the orchestration
+    # should still fail. But, number of times the sub-orchestrator and activity
+    # is called should increase as per the retry policies.
+
+    # First setup: No retry policies
```
cgillum, line 283: It seems weird to execute a test as part of `test_retry_policies` that doesn't have a retry policy. I suggest we break these into two separate tests, one which tests the default failure behavior and one which tests the behavior of retry policies.

tests/test_orchestration_executor.py
```diff
@@ -228,6 +228,173 @@ def orchestrator(ctx: task.OrchestrationContext, orchestrator_input):
...
+    # Simulate the task failing for the first time and confirm that a timer is scheduled for 1 second in the future
+    old_events = [
+        helpers.new_orchestrator_started_event(timestamp=current_timestamp),
+        helpers.new_execution_started_event(name, TEST_INSTANCE_ID, encoded_input=None),
+        helpers.new_task_scheduled_event(1, task.get_name(dummy_activity))]
+    expected_fire_at = current_timestamp + timedelta(seconds=1)
+    
+    new_events = [
+        helpers.new_orchestrator_started_event(timestamp=current_timestamp),
+        helpers.new_task_failed_event(1, Exception("Kah-BOOOOM!!!"))]
```
cgillum, line 262: Shouldn't this be `ValueError`?

tests/test_orchestration_executor.py
```diff
@@ -228,6 +228,173 @@ def orchestrator(ctx: task.OrchestrationContext, orchestrator_input):
...
+    executor = worker._OrchestrationExecutor(registry, TEST_LOGGER)
+    actions = executor.execute(TEST_INSTANCE_ID, old_events, new_events)
+    print(actions)
```
cgillum, line 265: Should we remove these `print(actions)` statements? I assume those were used for debugging?

durabletask/task.py
DeepanshuA: How is number of attempts different than number of retries?
Or is it a nomenclature suggestion?

durabletask/task.py
DeepanshuA: I have renamed it to `_start_time`, as it is basically the time at which it starts and is used to calculate the retry_expiration time.
And, instead of setting it with a setter (which was required), passed it while initializing RetryableTask, so it is initialized with required value.

durabletask/task.py
DeepanshuA: Removed it

durabletask/task.py
DeepanshuA: returned None and checked accordingly.

durabletask/task.py
DeepanshuA: Realized that complete method is not required separately in TimerTask, so removed it.

durabletask/worker.py
DeepanshuA: update to `TypeError` and updated this comment to "Unexpected sub-orchestration task type"

durabletask/task.py
DeepanshuA: I have now only one method to check all conditions for retry (`compute_next_delay`).
This method is removed.

durabletask/task.py
DeepanshuA: moved this logic to `compute_next_delay`

durabletask/worker.py
DeepanshuA: yeah, now it is there 2 times.

tests/test_orchestration_e2e.py
DeepanshuA: I have removed first setup i.e. without retry.
Other existing tests are supposed to cater / are catering to without retry cases.

tests/test_orchestration_executor.py
DeepanshuA: updated

tests/test_orchestration_executor.py
DeepanshuA: my bad, I have removed

DeepanshuA:
@cgillum Please review this PR. It has been a quite stretched PR time-wise.
I have incorporated review comments.

durabletask/worker.py
```diff
@@ -606,17 +671,31 @@ def process_event(self, ctx: _RuntimeOrchestrationContext, event: pb.HistoryEven
...
-                sub_orch_task.fail(
-                    f"Sub-orchestration task #{task_id} failed: {failedEvent.failureDetails.errorMessage}",
-                    failedEvent.failureDetails)
-                ctx.resume()
+                if isinstance(sub_orch_task, task.RetryableTask):
+                    if sub_orch_task._retry_policy is not None:
+                        next_delay = sub_orch_task.compute_next_delay()
+                        if next_delay == None:
+                            sub_orch_task.fail(
+                                f"Sub-orchestration task #{task_id} failed: {failedEvent.failureDetails.errorMessage}",
```
cgillum, line 679: One potential improvement we could add here is including the number of retries in this error message.

tests/test_orchestration_executor.py
```diff
@@ -228,6 +228,169 @@ def orchestrator(ctx: task.OrchestrationContext, orchestrator_input):
...
+    new_events = [
+        helpers.new_orchestrator_started_event(current_timestamp),
+        helpers.new_timer_fired_event(4, current_timestamp)]
+    executor = worker._OrchestrationExecutor(registry, TEST_LOGGER)
+    actions = executor.execute(TEST_INSTANCE_ID, old_events, new_events)
+    assert len(actions) == 4
+    assert actions[1].HasField("scheduleTask")
+    assert actions[1].id == 1
+
+    # # Simulate the task failing for a fourth time and confirm that a timer is scheduled for 8 seconds in the future
```
cgillum, line 332: Why `# #` instead of just `#` for these comments?

tests/test_orchestration_executor.py
```diff
@@ -793,6 +956,183 @@ def orchestrator(ctx: task.OrchestrationContext, _):
...
+    encoded_output = json.dumps(dummy_activity(None, "Seattle"))
+    new_events = [helpers.new_task_completed_event(2, encoded_output)]
+    executor = worker._OrchestrationExecutor(registry, TEST_LOGGER)
+    actions = executor.execute(TEST_INSTANCE_ID, old_events, new_events)
+    complete_action = get_and_validate_single_complete_orchestration_action(actions)
+    assert complete_action.orchestrationStatus == pb.ORCHESTRATION_STATUS_COMPLETED
+    assert complete_action.result.value == encoded_output
+
+def test_when_all_with_retry():
+    """Tests that a when_any pattern works correctly"""
```
cgillum, line 1040:
```suggestion
    """Tests that a when_all pattern works correctly"""
```

tests/test_orchestration_executor.py
```diff
@@ -793,6 +956,183 @@ def orchestrator(ctx: task.OrchestrationContext, _):
...
+        t1 = ctx.call_activity(dummy_activity, 
+                               retry_policy=task.RetryPolicy(
+                                            first_retry_interval=timedelta(seconds=2),
+                                            max_number_of_attempts=3,
+                                            backoff_coefficient=4,
+                                            max_retry_interval=timedelta(seconds=5),
+                                            retry_timeout=timedelta(seconds=50)),
+                               input="Tokyo")
+        t2 = ctx.call_activity(dummy_activity, input="Seattle")
+        winner = yield task.when_all([t1, t2])
```
cgillum, line 1056: This orchestration logic doesn't make sense. The result of `when_all` is not a "winner". Rather, it's a list containing the results of both `t1` and `t2`. This means taht your if/else check below will always execute the `else` block no matter what. Please rewrite this orchestration so that it correctly uses the `task.when_all` API.

cgillum (commented):
Changes are looking good. Just a few fairly minor things.

durabletask/worker.py
DeepanshuA: This is logged when it finally fails after Either 1. All retries have exhausted Or 2. Retry timed out.
This is not actually logged with each retry. 

I can add such logging at each attempt count but would need some new definition to have this understanding that what attempt number is it And, if retry has timed out, etc.
Do you think that is required at each retry/attempt?

tests/test_orchestration_executor.py
DeepanshuA: Yeah, updated to return as per `when_all`

tests/test_orchestration_executor.py
DeepanshuA: My bad, copy-paste mistake

tests/test_orchestration_executor.py
DeepanshuA: Updated

durabletask/worker.py
cgillum: I'm not suggesting we log on each retry (though we could consider this too). I'm simply suggesting that the error message could include the number of retries that were attempted. For example,
```suggestion
                                f"Sub-orchestration task #{task_id} failed after {attempt_count} attempt(s): {failedEvent.failureDetails.errorMessage}",
```

cgillum (approved):

PR #13, opened 2023-08-14
Fix Activity Input check

DeepanshuA:
This PR fixes issue https://github.com/dapr/python-sdk/issues/574

0 in Activity Input won't be wrongly converted to None.


cgillum (commented):
The change looks good, but can we add a regression test to 1) prove the fix works (it should fail if your change is reverted) and 2) make sure we don't hit it again?

cgillum:
Ping @DeepanshuA

PR #14, opened 2023-08-14
Remove Python 3.7 support due to EOL

berndverst:
Python 3.7 is now end of life and no longer receives security updates.

This removes Python 3.7 support and also removes it from CI runs.

(The next release of the Dapr SDK has also removed Python 3.7 support)

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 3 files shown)


berndverst:
cc @DeepanshuA

berndverst:
As an aside, if you to end up using asyncio more, in Python 3.8+ there are convenient async unittest tools. In fact, `asyncio` only has a stable API starting in 3.8. The test helpers do not exist in Python 3.7 - so that's another reason to switch to 3.8 as the minimum version now.

Some other cool additions in Python 3.8: https://docs.python.org/3/whatsnew/3.8.html

cgillum (approved):
Thanks for this PR!

PR #16, opened 2023-09-13
Add metadata option to intercept grpc channel

DeepanshuA:
This PR adds option to pass a metadata to be used to intercept grpc channel.
This can be used to pass API Token kind of information.

### DIFF
--- a/durabletask/internal/grpc_interceptor.py
+++ b/durabletask/internal/grpc_interceptor.py
@@ -1,56 +1,91 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from typing import Any, Dict, Optional
+from collections import namedtuple
+from typing import List, Tuple
 
 import grpc
 
+
+class _ClientCallDetails(
+        namedtuple(
+            '_ClientCallDetails',
+            ['method', 'timeout', 'metadata', 'credentials', 'wait_for_ready', 'compression']),
+        grpc.ClientCallDetails):
+    """This is an implementation of the ClientCallDetails interface needed for interceptors.
+    This class takes six named values and inherits the ClientCallDetails from grpc package.
+    This class encloses the values that describe a RPC to be invoked.
+    """
+    pass
+
+def intercept_call(
+            client_call_details: _ClientCallDetails, metadata: List[Tuple[str, str]]) -> _ClientCallDetails:
+    new_metadata = []
+    new_client_call_details = client_call_details
+    if client_call_details.metadata is not None:
+        new_metadata = list(metadata)
+    if metadata is not None:
+        new_metadata.extend(metadata)
+        new_client_call_details = _ClientCallDetails(
+            client_call_details.method,
+            client_call_details.timeout,
+            metadata=new_metadata,
+            credentials=client_call_details.credentials,
+            wait_for_ready=client_call_details.wait_for_ready,
+            compression=client_call_details.compression,
+        )
+    return new_client_call_details
 
 class UnaryUnaryClientInterceptorImpl (grpc.UnaryUnaryClientInterceptor):
     """The class implements a UnaryUnaryClientInterceptor from grpc to add an interceptor to add
     additional headers to all calls as needed."""
 
-    def __init__(self, metadata: Optional[Dict[str, Any]]):
+    def __init__(self, metadata: List[Tuple[str, str]]):
         super().__init__()
         self._metadata = metadata
 
     def intercept_unary_unary(self, continuation, client_call_details, request):
-        if self._metadata:
-            client_call_details = client_call_details._replace(metadata=self._metadata)
-        return continuation(client_call_details, request)
+        new_client_call_details = intercept_call(client_call_details, self._metadata)
+        return continuation(new_client_call_details, request)
 
 class UnaryStreamClientInterceptorImpl (grpc.UnaryStreamClientInterceptor):
     """The class implements a UnaryStreamClientInterceptor from grpc to add an interceptor to add
     additional headers to all calls as needed."""
 
-    def __init__(self, metadata: Optional[Dict[str, Any]]):
+    def __init__(self, metadata: List[Tuple[str, str]]):
         super().__init__()
         self._metadata = metadata
 
     def intercept_unary_stream(self, continuation, client_call_details, request):
-        client_call_details = client_call_details._replace(metadata=self._metadata)
-        return continuation(client_call_details, request)
+        # client_call_details = client_call_details._replace(metadata=self._metadata)
+        # return continuation(client_call_details, request)
+        new_client_call_details = intercept_call(client_call_details, self._metadata)
+        return continuation(new_client_call_details, request)
 
 class StreamUnaryClientInterceptorImpl (grpc.StreamUnaryClientInterceptor):
     """The class implements a StreamUnaryClientInterceptor from grpc to add an interceptor to add
     additional headers to all calls as needed."""
 
-    def __init__(self, metadata: Optional[Dict[str, Any]]):
+    def __init__(self, metadata: List[Tuple[str, str]]):
         super().__init__()
         self._metadata = metadata
 
     def intercept_stream_unary(self, continuation, client_call_details, request):
-        client_call_details = client_call_details._replace(metadata=self._metadata)
-        return continuation(client_call_details, request)
+        # client_call_details = client_call_details._replace(metadata=self._metadata)
+        # return continuation(client_call_details, request)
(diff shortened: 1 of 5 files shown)


durabletask/internal/grpc_interceptor.py
```diff
@@ -0,0 +1,56 @@
...
+    """The class implements a UnaryUnaryClientInterceptor from grpc to add an interceptor to add
+    additional headers to all calls as needed."""
+
+    def __init__(self, metadata: Optional[Dict[str, Any]]):
+        super().__init__()
+        self._metadata = metadata
+
+    def intercept_unary_unary(self, continuation, client_call_details, request):
+        if self._metadata:
+            client_call_details = client_call_details._replace(metadata=self._metadata)
```
cgillum, line 19: Why are we using a private method here? And what does it do?

cgillum (changes requested):

durabletask/internal/grpc_interceptor.py
DeepanshuA: That was for updating metadata in client_call_details.
I have updated that now.

durabletask/internal/grpc_interceptor.py
```diff
@@ -0,0 +1,93 @@
...
+            client_call_details: _ClientCallDetails, metadata: List[Tuple[str, str]]) -> _ClientCallDetails:
+    """Internal intercept_call implementation which adds metadata to grpc metadata in the RPC
+        call details."""
+    new_metadata = []
+    new_client_call_details = client_call_details
+    if client_call_details.metadata is not None:
+        new_metadata = list(metadata)
+    if metadata is not None:
+        new_metadata.extend(metadata)
+        new_client_call_details = _ClientCallDetails(
```
cgillum, line 31: Why do we need to create a new copy of the `_ClientCallDetails` object? Can't we mutate the `client_call_details` variable that's passed in?

cgillum (commented):
The approach of using an interceptor makes sense for this use-case. However, I'm wondering if we can reduce the overhead of the current implementation and make it a bit simpler. See my code comments.

durabletask/worker.py
```diff
@@ -85,10 +86,12 @@ class TaskHubGrpcWorker:
 
     def __init__(self, *,
                  host_address: Union[str, None] = None,
+                 metadata: Optional[Dict[str, Any]],
```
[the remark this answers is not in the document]
DeepanshuA: Left by mistake, corrected that.

durabletask/internal/grpc_interceptor.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from collections import namedtuple
+from typing import List, Tuple
+
+import grpc
+
+
+class _ClientCallDetails(
```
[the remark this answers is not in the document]
DeepanshuA: I have used single implementation for all interceptors now.
Also, would need to create an implementation for grpc.ClientCallDetails interface and also create new `metadata` object.
As, client_call_details.metadata is otherwise not accessible in required way.
Also, client_call_details.metadata is read-only.

image

Please refer: https://github.com/grpc/grpc/blob/62521a889fe134859aed7bdf4ea0730534f15d09/examples/python/interceptors/headers/header_manipulator_client_interceptor.py#L38

https://stackoverflow.com/a/54743669/18848251

durabletask/internal/grpc_interceptor.py
```diff
@@ -0,0 +1,93 @@
...
+    This class encloses the values that describe a RPC to be invoked.
+    """
+    pass
+
+def _intercept_call(
+            client_call_details: _ClientCallDetails, metadata: List[Tuple[str, str]]) -> _ClientCallDetails:
+    """Internal intercept_call implementation which adds metadata to grpc metadata in the RPC
+        call details."""
+    new_metadata = []
+    new_client_call_details = client_call_details
```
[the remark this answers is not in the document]
DeepanshuA: Modifying client_call_details only in _intercept_call, but need to create a new `_ClientCallDetails`. Please refer  https://github.com/microsoft/durabletask-python/pull/16#discussion_r1330163140

durabletask/internal/grpc_interceptor.py
```diff
@@ -0,0 +1,93 @@
...
+    This class takes six named values and inherits the ClientCallDetails from grpc package.
+    This class encloses the values that describe a RPC to be invoked.
+    """
+    pass
+
+def _intercept_call(
+            client_call_details: _ClientCallDetails, metadata: List[Tuple[str, str]]) -> _ClientCallDetails:
+    """Internal intercept_call implementation which adds metadata to grpc metadata in the RPC
+        call details."""
+    new_metadata = []
```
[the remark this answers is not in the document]
DeepanshuA: Please refer https://github.com/microsoft/durabletask-python/pull/16#discussion_r1330163140

durabletask/internal/grpc_interceptor.py
DeepanshuA: Please refer https://github.com/microsoft/durabletask-python/pull/16#discussion_r1330163140

durabletask/internal/grpc_interceptor.py
```diff
@@ -18,76 +18,44 @@ class _ClientCallDetails(
...
         self._metadata = metadata
 
-    def intercept_unary_unary(self, continuation, client_call_details, request):
-        new_client_call_details = _intercept_call(client_call_details, self._metadata)
-        return continuation(new_client_call_details, request)
+    def _intercept_call(
+                self, client_call_details: _ClientCallDetails) -> grpc.ClientCallDetails:
+        """Internal intercept_call implementation which adds metadata to grpc metadata in the RPC
+            call details."""
+        metadata = []
```
cgillum, line 37: Let's move this into an `else:` block so that we don't make any unnecessary allocations.

durabletask/internal/shared.py
```diff
@@ -18,13 +20,15 @@ def get_default_host_address() -> str:
     return "localhost:4001"
 
 
-def get_grpc_channel(host_address: Union[str, None]) -> grpc.Channel:
+def get_grpc_channel(host_address: Union[str, None], metadata: List[Tuple[str, str]]) -> grpc.Channel:
```
cgillum, line 23: Should this be optional?
```suggestion
def get_grpc_channel(host_address: Union[str, None], metadata: Union[List[Tuple[str, str]], None]) -> grpc.Channel:
```

durabletask/internal/shared.py
```diff
@@ -18,13 +20,15 @@ def get_default_host_address() -> str:
...
     if host_address is None:
         host_address = get_default_host_address()
     channel = grpc.insecure_channel(host_address)
+    if metadata != None and len(metadata) > 0:
```
cgillum, line 27: nit: I don't think we use `!= None` much in this codebase.
```suggestion
    if metadata is not None and len(metadata) > 0:
```

cgillum (approved):
Looks good. Just a few nits.

durabletask/internal/shared.py
DeepanshuA: addressed

durabletask/internal/shared.py
DeepanshuA: yeah, made it optional

durabletask/internal/grpc_interceptor.py
DeepanshuA: I moved that to an `else` block and also moved creation of _ClientCallDetails, only if passed metadata is not None.

durabletask/internal/grpc_interceptor.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def _intercept_call(
+                self, client_call_details: _ClientCallDetails) -> grpc.ClientCallDetails:
+        """Internal intercept_call implementation which adds metadata to grpc metadata in the RPC
+            call details."""
+        if client_call_details.metadata is not None:
+            metadata = list(client_call_details.metadata)
+        else:
+            metadata = []
+        if self._metadata is not None:
```
cgillum, line 41: This check should be moved to the very beginning of the function. We don't want to allocate anything if `self._metadata` is not initialized.

```python
def _intercept_call(...):
    if self._metadata is None:
        return client_call_details

    if client_call_details.metadata is not None:
        ...
```

cgillum (commented):
Last comment, again for performance.

durabletask/internal/grpc_interceptor.py
DeepanshuA: incorporated

PR #31, opened 2024-10-24
update durabletask protos, set custom status

famarting:
initial work for https://github.com/dapr/dapr/issues/8008#issuecomment-2404979325

here I updated the durabletask protobuf definitions and regenerated the client code.

Also added support for https://github.com/dapr/python-sdk/issues/739 with a new test included

Also started https://github.com/dapr/python-sdk/issues/655 but have not added a test yet


famarting:
I wanted to create this PR first and check with @cgillum if everything looks ok... the grpc code that got generated seems a bit weird to me

durabletask/task.py
```diff
@@ -71,6 +71,20 @@ def is_replaying(self) -> bool:
...
+    @property
+    @abstractmethod
+    def custom_status(self) -> str:
+        """Get the custom status.
+        """
+        pass
+
+    @custom_status.setter
+    @abstractmethod
+    def custom_status(self, custom_status: str) -> None:
```
cgillum, line 83: Can we do `set_custom_status` instead of a `custom_status` property setter? This is how our other SDKs work (using a method rather than a property), and gives us greater flexibility in the future if we want to add parameters or make behavior changes that wouldn't be natural with a property.

durabletask/task.py
```diff
@@ -71,6 +71,20 @@ def is_replaying(self) -> bool:
         """
         pass
 
+    @property
+    @abstractmethod
+    def custom_status(self) -> str:
+        """Get the custom status.
```
cgillum, line 77: I don't think we need a getter for this. We treat custom status as "write-only" in other SDKs since there shouldn't generally be a need for reading the custom status in this way (you can use local variables instead if you want to remember it).

tests/test_orchestration_e2e.py
```diff
@@ -441,3 +441,26 @@ def throw_activity(ctx: task.ActivityContext, _):
...
+        state = c.wait_for_orchestration_completion(id, timeout=30)
+
+    assert state is not None
+    assert state.name == task.get_name(empty_orchestrator)
+    assert state.instance_id == id
+    assert state.failure_details is None
+    assert state.runtime_status == client.OrchestrationStatus.COMPLETED
+    assert state.serialized_input is None
+    assert state.serialized_output is None
+    assert state.serialized_custom_status is "foobaz"
```
cgillum, line 466: The expectation in this case should be that `serialized_custom_status` is `"\"foobaz\""` (a serialized value), not `"foobaz"`. See the other tests were we do asserts against `serialized_input` and `serialized_output`.

cgillum (changes requested):
Left a few requests for changes.

I haven't looked at the auto-generated protobuf changes closely, but I appreciate that we're updating the submodule reference. Will run this through the CI to see if it causes any issues.

durabletask/task.py
famarting: done

durabletask/task.py
famarting: done

famarting:
@cgillum I just noticed this comment https://github.com/microsoft/durabletask-python/blob/main/Makefile#L14 which could explain the unexpected build errors I'm getting on CI...

do you know what changes are missing to the proto generated files?

famarting:
found this https://github.com/microsoft/durabletask-python/blob/main/durabletask/internal/orchestrator_service_pb2_grpc.py#L7 , but I think I still have test failures somehow due to the generated code

cgillum:
This seems like a different type of error than what I’d expect in the case where I needed to update an import statement. I can try to help take a look but I may not be able to get to this right away due to some other urgent priorities at the moment.

famarting:
> This seems like a different type of error than what I’d expect in the case where I needed to update an import statement. I can try to help take a look but I may not be able to get to this right away due to some other urgent priorities at the moment.

ok, it was much simpler than I anticipated, actually after fixing the protobuf generation it was just a matter of fixing the build on the tests

Issue #8, opened 2023-04-28
Protos, Grpcio-tool and protobuf version incompatibilities


berndverst:
Note that protobuf>=4.22 introduced a breaking change, and protos generated using grpcio-tools>=1.49 are only compatible with protbuf>=4.22

Most ML libraries however force `protobuf<4.22` still. That's why Dapr too has pinned `protobuf<4.22`

I recommend pinning `grpcio-tools` here to `<1.49`. There is however no need to pin `grpcio` itself - any version is fine.

berndverst:
Fixed in #36.

We are pinning `grpcio-tools=1.62.3` which under the hood uses `protobuf==3.25.X` so that should work just fine with the ML libraries.

The minimum `grpcio` version that can definitely load those compiled protos is `1.60.0`. The user can decide which `grpc` version to use.

berndverst:
Strangely I cannot find any `grpcio-tools` versions that use protobuf 4.X anymore at all. Either way, with the changes there should not be any further problems.

PR #36, opened 2025-01-08
Downgrade required `grpcio` and `protobuf` versions

berndverst:
This downgrades the required `grpcio` version by pinning `grpcio-tools` as dev-only requirement to version 1.62.3 which is the last release before a breaking change in protobuf is introduced (with 1.26.X / 3.26.X).

The protos are recompiled with this version. (Note: protos compiled with grpcio-tools 1.67.0 for example MUST be loaded with grpcio 1.67.0 or higher hence recompilation with an older version was necessary)

Additional improvements:
The proto generation has been changed to no longer require manual edits.


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `set_custom_status` orchestrator API ([#31](https://github.com/microsoft/durabletask-python/pull/31)) - contributed by [@famarting](https://github.com/famarting)
 - Added `purge_orchestration` client API ([#34](https://github.com/microsoft/durabletask-python/pull/34)) - contributed by [@famarting](https://github.com/famarting)
 
+### Changes
+
+- Protos are compiled with gRPC 1.62.3 / protobuf 3.25.X instead of the latst release. This ensures compatibility with a wider range of grpcio versions for better compatibility with other packages / libraries.
```
berndverst, line 17:
```suggestion
- Protos are compiled with gRPC 1.62.3 / protobuf 3.25.X instead of the latest release. This ensures compatibility with a wider range of grpcio versions for better compatibility with other packages / libraries.
```

berndverst:
This fixes https://github.com/microsoft/durabletask-python/issues/8

PR #38, opened 2025-01-10
Remove protocol prefix from host name and auto-configure secure mode

berndverst:
Remove protocol prefixes from host name and auto-configure secure mode

This addresses an issue where connections fail if the provided host name starts with a `http://` or `https://` prefix.

When `https` and `grpcs` were provided as prefixes, we force secure mode to be enabled.


cgillum (approved):
LGTM!

PR #40, opened 2025-01-22
Creation of DTS example and passing of completionToken

RyanLettieri:
This PR adds in the completion token to the worker.py class for DTS to understand when a task or work item is complete.

This PR also adds in an example of using the DTS backend and updates the README to reflect the necessary steps


examples/README.md
```diff
@@ -1,14 +1,39 @@
 # Examples
 
-This directory contains examples of how to author durable orchestrations using the Durable Task Python SDK.
+This directory contains examples of how to author durable orchestrations using the Durable Task Python SDK. There are two backends that are compatible with the Durable Task Python SDK: The Dapr sidecar, and the Durable Task Scheduler (DTS)
```
cgillum, line 3: nit: complete the sentence with a period.

examples/README.md
```diff
@@ -1,14 +1,39 @@
...
 
 All the examples assume that you have a Durable Task-compatible sidecar running locally. There are two options for this:
 
 1. Install the latest version of the [Dapr CLI](https://docs.dapr.io/getting-started/install-dapr-cli/), which contains and exposes an embedded version of the Durable Task engine. The setup process (which requires Docker) will configure the workflow engine to store state in a local Redis container.
 
-1. Clone and run the [Durable Task Sidecar](https://github.com/microsoft/durabletask-go) project locally (requires Go 1.18 or higher). Orchestration state will be stored in a local sqlite database.
+2. Clone and run the [Durable Task Sidecar](https://github.com/microsoft/durabletask-go) project locally (requires Go 1.18 or higher). Orchestration state will be stored in a local sqlite database.
+
+
+## Prerequisites for using DTS
```
cgillum, line 14: These instructions apply only to one of the several examples. I suggest we put DTS-specific instructions and DTS-specific samples in a sub-directory under `examples`. For example, `/examples/dts/activity_sequence_dts.py`, etc.

examples/dts_activity_sequence.py
```diff
@@ -0,0 +1,73 @@
...
+
+# Read the environment variable
+endpoint = os.getenv("ENDPOINT")
+
+# Check if the variable exists
+if endpoint:
+    print(f"The value of ENDPOINT is: {endpoint}")
+else:
+    print("ENDPOINT is not set. Please set the ENDPOINT environment variable to the endpoint of the taskhub")
+    print("If you are using windows powershell, run the following: $env:ENDPOINT=\"<taskhubEndpoint>\"")
```
cgillum, line 45: There's no such thing as a task hub endpoint. Only a scheduler endpoint.

examples/dts_activity_sequence.py
```diff
@@ -0,0 +1,73 @@
...
+
+default_credential = DefaultAzureCredential()
+# Define the scope for Azure Resource Manager (ARM)
+arm_scope = "https://durabletask.io/.default"
+
+# Retrieve the access token
+access_token = "Bearer " + default_credential.get_token(arm_scope).token
+# create a client, start an orchestration, and wait for it to finish
+metaData: list[tuple[str, str]] = [
+    ("taskhub", taskhub_name), # Hardcode for now, just the taskhub name
```
cgillum, line 58: Does this comment about hardcoding need to be removed?

examples/dts_activity_sequence.py
```diff
@@ -0,0 +1,73 @@
...
+    print("If you are using bash, run the following: export ENDPOINT=\"<taskhubEndpoint>\"")
+    exit()
+
+
+default_credential = DefaultAzureCredential()
+# Define the scope for Azure Resource Manager (ARM)
+arm_scope = "https://durabletask.io/.default"
+
+# Retrieve the access token
+access_token = "Bearer " + default_credential.get_token(arm_scope).token
```
cgillum, line 55: We should add a caveat in the comments that this approach doesn't support token refresh and thus can't be used in production.

examples/README.md
```diff
@@ -1,14 +1,39 @@
...
+
+All the examples assume that you have a Durable Task Scheduler taskhub created.
+
+The simplest way to create a taskhub is by using the az cli commands:
+
+1. Create a scheduler:
+    az durabletask scheduler create --resource-group <testrg> --name <testscheduler> --location <eastus> --ip-allowlist "[0.0.0.0/0]" --sku-capacity 1, --sku-name "Dedicated" --tags "{}"
```
cgillum, line 21: Remove the comma.
```suggestion
    az durabletask scheduler create --resource-group <testrg> --name <testscheduler> --location <eastus> --ip-allowlist "[0.0.0.0/0]" --sku-capacity 1 --sku-name "Dedicated" --tags "{}"
```

cgillum (commented):
The code changes look good, though I have feedback on the sample and README.

berndverst (approved):

examples/dts/dts_activity_sequence.py
```diff
@@ -0,0 +1,73 @@
...
+    print("If you are using bash, run the following: export ENDPOINT=\"<schedulerEndpoint>\"")
+    exit()
+
+
+default_credential = DefaultAzureCredential()
+# Define the scope for Azure Resource Manager (ARM)
+arm_scope = "https://durabletask.io/.default"
+
+# Retrieve the access token. Note that this approach doesn't support token refresh and can't be used in production.
+access_token = "Bearer " + default_credential.get_token(arm_scope).token
```
torosent, line 55: The missing refresh token handling makes this sample less practical. Please add an implementation that supports refresh tokens. For instance, a `TokenProvider` that utilizes the [various credential classes](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity?view=azure-python) to ensure the token is always valid by handling expiration and refreshing.

examples/dts/dts_activity_sequence.py
```diff
@@ -0,0 +1,73 @@
...
+    w.add_activity(hello)
+    w.start()
+
+    c = client.TaskHubGrpcClient(host_address=endpoint, metadata=metaData, secure_channel=True)
+    instance_id = c.schedule_new_orchestration(sequence)
+    state = c.wait_for_orchestration_completion(instance_id, timeout=45)
+    if state and state.runtime_status == client.OrchestrationStatus.COMPLETED:
+        print(f'Orchestration completed! Result: {state.serialized_output}')
+    elif state:
+        print(f'Orchestration failed: {state.failure_details}')
```
torosent, line 73: Please add comments to each line to explain its functionality.

torosent (changes requested):
`azure-identity` is missing from `requirements.txt`

examples/dts/dts_activity_sequence.py
berndverst: I think that is fair - but rather than putting this into the sample (making the sample look unnecessarily complex), why don't we work on a better high level client that wraps all of this seamlessly. The client can take an azure identity TokenCredential as input parameter. If none is provided it can use the DefaultCredential.

My vote would be for doing a follow up PR for a high level client and then updating the DTS samples to use the high level client.

durabletask/worker.py
```diff
@@ -88,14 +89,36 @@ def __init__(self, *,
...
+            updated = False
+            for i, (key, _) in enumerate(self._metadata):
+                if key == "authorization":
+                    self._metadata[i] = ("authorization", token)
+                    updated = True
+                    break
+            
+            # If not updated, add a new entry
+            if not updated:
+                self._metadata.append(("authorization", token))
```
cgillum, line 121: The way we're building these SDKs is that the core functionality is general purpose and applicable to a variety of different backend implementations. For example, Dapr, DTS, or something else. However, this change here is specific to DTS (because it assumes an "authorization" header that requires a token) and therefore violates the design principle we're going for.  I would ask that you follow the example of the .NET SDK and create a separate package which is specific to DTS that is able to extend the core worker logic. It's okay if we also need to change `worker.py`, but only if it's a general purpose change that could potentially benefit multiple backends.

examples/dts/dts_activity_sequence.py
```diff
@@ -0,0 +1,71 @@
...
+]
+
+# configure and start the worker
+with worker.TaskHubGrpcWorker(host_address=endpoint, metadata=metaData, secure_channel=True, access_token_manager=token_manager) as w:
+    w.add_orchestrator(sequence)
+    w.add_activity(hello)
+    w.start()
+
+    # Construct the client and run the orchestrations
+    c = client.TaskHubGrpcClient(host_address=endpoint, metadata=metaData, secure_channel=True)
```
cgillum, line 65: The client also needs an ability to refresh tokens.

examples/dts/dts_activity_sequence.py
```diff
@@ -0,0 +1,71 @@
+import os
+from azure.identity import DefaultAzureCredential
+
+"""End-to-end sample that demonstrates how to configure an orchestrator
+that calls an activity function in a sequence and prints the outputs."""
+from durabletask import client, task, worker
+from durabletask.accessTokenManager import AccessTokenManager
```
cgillum, line 7: Per Python naming conventions, we should use snake_case in module names rather than camelCase. For example, `access_token_manager` instead of `accessTokenManager`. Please see how we're naming all the other files in this repo.

cgillum (changes requested):

durabletask/worker.py
berndverst: @cgillum instead of `durabletask.client` and `durabletask.worker` would having `durabletask.dts.clients` and `durabletask.dts.worker` work? Those of course should utilize the former in their implementation - and then any Azure credential, token refresh and other logic can be added as well.

cgillum: Yeah, that's roughly what I was hoping we could do. So instead of a developer creating a `TaskHubGrpcWorker/Client` and passing in a bunch of generic parameters, they'd create a `DurableTaskSchedulerWorker/Client` defined in a new `durabletask-azuremanaged` package from PyPI, which has constructors that are specifically designed for use with DTS. Internally, they'd use the `durabletask.client` and `durabletask.worker` as you said.

berndverst: So you don't want it as a subpackage in the main `durabetask-python` library, but instead publish it as an entirely different package?

I guess that makes sense because we could avoid dependency bloat (for those who don't want to install Azure SDK dependencies like `azure-identity) when using durable task. That's the biggest advantage of the separate package.

I'll let you and @RyanLettieri decide - I can help review.

cgillum: Correct. This should be an entirely different package to avoid the dependency bloat. It's the same strategy we're using for the .NET and Java SDKs.

examples/dts/dts_activity_sequence.py
```diff
@@ -0,0 +1,71 @@
...
+    print(f"The value of ENDPOINT is: {endpoint}")
+else:
+    print("ENDPOINT is not set. Please set the ENDPOINT environment variable to the endpoint of the scheduler")
+    print("If you are using windows powershell, run the following: $env:ENDPOINT=\"<schedulerEndpoint>\"")
+    print("If you are using bash, run the following: export ENDPOINT=\"<schedulerEndpoint>\"")
+    exit()
+
+
+# Define the scope for Azure Resource Manager (ARM)
+arm_scope = "https://durabletask.io/.default"
```
torosent, line 51: Should this be in the Token manager? Why whould a customer define something different fot DTS?

examples/dts/dts_activity_sequence.py
```diff
@@ -52,18 +51,18 @@ def sequence(ctx: task.OrchestrationContext, _):
 arm_scope = "https://durabletask.io/.default"
 token_manager = AccessTokenManager(scope = arm_scope)
 
-metaData: list[tuple[str, str]] = [
+meta_data: list[tuple[str, str]] = [
```
cgillum, line 54: The fact that we internally use the metadata dictionary to pass task hub information is an implementation detail that developers should not need to know or care about. This dictionary is intended to be used in cases where the developer needs to do something custom that isn't natively understood by the SDK. Task hubs do not fit into this bucket, because they are a first class concept in DTS. We should instead allow the developer to pass the task hub name directly in the DurableTaskSchedulerWorker and client class constructors in a type-safe way via a typed `taskhub: str` parameter that doesn't require them to look at documentation.

examples/dts/dts_activity_sequence.py
```diff
@@ -52,18 +51,18 @@ def sequence(ctx: task.OrchestrationContext, _):
...
 # configure and start the worker
-with DurableTaskSchedulerWorker(host_address=endpoint, metadata=metaData, secure_channel=True, access_token_manager=token_manager) as w:
+with DurableTaskSchedulerWorker(host_address=endpoint, metadata=meta_data, secure_channel=True, access_token_manager=token_manager) as w:
     w.add_orchestrator(sequence)
     w.add_activity(hello)
     w.start()
 
     # Construct the client and run the orchestrations
-    c = client.TaskHubGrpcClient(host_address=endpoint, metadata=metaData, secure_channel=True)
+    c = DurableTaskSchedulerClient(host_address=endpoint, metadata=meta_data, secure_channel=True, access_token_manager=token_manager)
```
cgillum, line 65: Similar feedback about token manager. Developers shouldn't have to worry about creating or configuring something like this because it's a lower-level detail unrelated to their business logic. Instead, the DurableTaskSchedulerClient and worker types should create an access token manager _internally_ and not make it a responsibility of the developer.

durabletask/internal/shared.py
```diff
@@ -10,6 +10,7 @@
 import grpc
 
 from durabletask.internal.grpc_interceptor import DefaultClientInterceptorImpl
+from externalpackages.durabletaskscheduler.durabletask_grpc_interceptor import DTSDefaultClientInterceptorImpl
```
cgillum, line 13: We should not have DTS-specific code in the `durabletask` library. If we need new abstractions in the `durabletask` library, then we need to make them generic in such a way that they can be used for DTS or for non-DTS scenarios.

durabletask/client.py
```diff
@@ -97,6 +97,7 @@ def __init__(self, *,
                  log_handler: Optional[logging.Handler] = None,
                  log_formatter: Optional[logging.Formatter] = None,
                  secure_channel: bool = False):
+        self._metadata = metadata
```
cgillum, line 100: Is this change needed? Seems like we're not using `_metadata` anywhere in this file.

examples/dts/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+```sh
+export ENDPOINT=<taskhubEndpoint>
+```
+
+5. Since the samples rely on azure identity, ensure the package is installed and up-to-date
+
+```sh
+python3 -m pip install azure-identity
```
cgillum, line 32: Shouldn't there be a DTS-specific package that users need to install? For example, I was expecting something like `durabletask-azuremanaged` as being a required module to install, and it would include `azure-identity` as one of its implicit dependencies.

cgillum: Also, shouldn't there be a requirements.txt file for these examples?

cgillum (changes requested):
Haven't finished reviewing the whole thing yet, but have some fairly significant feedback for what I've reviewed so far.

examples/dts/README.md
RyanLettieri: That package will be installed when the user runs `python3` -m pip install .` as suggested in the higher level README. I've added the steps to this README as well

durabletask/client.py
```diff
@@ -96,8 +98,23 @@ def __init__(self, *,
                  metadata: Optional[list[tuple[str, str]]] = None,
                  log_handler: Optional[logging.Handler] = None,
                  log_formatter: Optional[logging.Formatter] = None,
-                 secure_channel: bool = False):
-        channel = shared.get_grpc_channel(host_address, metadata, secure_channel=secure_channel)
+                 secure_channel: bool = False,
+                 interceptors: Optional[list] = None):
```
cgillum, line 102: Can we further constrain the type of objects we allow in this `list`, like we do in `worker.py`?

durabletask/worker.py
```diff
@@ -88,15 +90,25 @@ def __init__(self, *,
                  metadata: Optional[list[tuple[str, str]]] = None,
                  log_handler=None,
                  log_formatter: Optional[logging.Formatter] = None,
-                 secure_channel: bool = False):
+                 secure_channel: bool = False,
+                 interceptors: Optional[list[grpc.ServerInterceptor]] = None):  # Add interceptors
```
cgillum, line 94: This might be just my ignorance on gRPC + Python, but `grpc.ServerInterceptor` sounds like something you'd put on the server, whereas the worker is more like a client. Is this the right type for us to be using or in Python is a `ServerInterceptor` also used for clients?

durabletask/worker.py
```diff
@@ -88,15 +90,25 @@ def __init__(self, *,
...
         self._registry = _Registry()
         self._host_address = host_address if host_address else shared.get_default_host_address()
-        self._metadata = metadata
+        self._metadata = metadata or []  # Ensure metadata is never None
```
cgillum, line 97: I don't think we have this same check in client.py. Should we? Or do we even need this?

durabletask/worker.py
```diff
@@ -88,15 +90,25 @@ def __init__(self, *,
...
         self._logger = shared.get_logger("worker", log_handler, log_formatter)
         self._shutdown = Event()
         self._is_running = False
         self._secure_channel = secure_channel
 
+        # Determine the interceptors to use
+        if interceptors is not None:
+            self._interceptors = interceptors
+        elif self._metadata:
+            self._interceptors = [DefaultClientInterceptorImpl(self._metadata)]
```
cgillum, line 107: If someone provides both `interceptors` and `metadata`, then we probably want to do both the `DefaultClientInterceptorImpl` and also whatever interceptors that the caller provided, right? Otherwise we'd be totally ignoring the value they provided for `metadata`.

durabletask/worker.py
```diff
@@ -117,7 +129,12 @@ def add_activity(self, fn: task.Activity) -> str:
 
     def start(self):
         """Starts the worker on a background thread and begins listening for work items."""
-        channel = shared.get_grpc_channel(self._host_address, self._metadata, self._secure_channel)
+        
+        if self._metadata:
+            interceptors = [DefaultClientInterceptorImpl(self._metadata)]  
+        else:
+            interceptors =  None
```
cgillum, line 136: We already initialized the interceptors list as part of `self._interceptors` in the constructor, so it seems like we shouldn't be doing this here, and below we should be passing in `self._interceptors` to the `shared.get_grpc_channel` call.

durabletask/internal/shared.py
```diff
@@ -25,8 +23,10 @@ def get_default_host_address() -> str:
 
 def get_grpc_channel(
         host_address: Optional[str],
-        metadata: Optional[list[tuple[str, str]]],
-        secure_channel: bool = False) -> grpc.Channel:
+        metadata: Optional[list[tuple[str, str]]] = None,
+        secure_channel: bool = False,
+        interceptors: Optional[list] = None) -> grpc.Channel:
```
cgillum, line 28: Same question about whether we should be further constraining the allowed types inside of this `list`.

cgillum (changes requested):
The updated design is looking great. A few technical things that we still need to clean up. I haven't reviewed everything quite yet, though.

durabletask/client.py
```diff
@@ -96,8 +98,25 @@ def __init__(self, *,
                  metadata: Optional[list[tuple[str, str]]] = None,
                  log_handler: Optional[logging.Handler] = None,
                  log_formatter: Optional[logging.Formatter] = None,
-                 secure_channel: bool = False):
-        channel = shared.get_grpc_channel(host_address, metadata, secure_channel=secure_channel)
+                 secure_channel: bool = False,
+                 interceptors: Optional[list[DefaultClientInterceptorImpl]] = None):
```
cgillum, line 102: This doesn't look right. `DefaultClientInterceptorImpl` is an internal class that we own and shouldn't be exposing to our users. Rather, there should be some base class specified here, like `grpc.UnaryUnaryClientInterceptor` or `grpc.UnaryStreamClientInterceptor`, that customers can derive their own interceptor types from. So this should be something like:
```suggestion
                 interceptors: Optional[list[Union[grpc.UnaryUnaryClientInterceptor, grpc.UnaryStreamClientInterceptor]]] = None):
```

cgillum (commented):
We're getting much closer. Glad we figured out the package stuff. Now I have a bunch more feedback on the API design and some of the internal implementation details. I haven't looked closely at everything yet but feel like I reached a good stopping poitn.

examples/dts/dts_activity_sequence.py
```diff
@@ -2,8 +2,8 @@
 that calls an activity function in a sequence and prints the outputs."""
 import os
 from durabletask import client, task
```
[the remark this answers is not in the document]
RyanLettieri: We use this import for comparing against the orchestration status since the regular client class contains the OrchestrationStatus enum. I can import it into the base class and use it in the examples if you'd like

durabletask-azuremanaged/durabletask/azuremanaged/client.py
```diff
@@ -14,8 +14,7 @@ def __init__(self, *,
                  taskhub: str,
                  secure_channel: Optional[bool] = True,
                  metadata: Optional[list[tuple[str, str]]] = None,
-                 use_managed_identity: Optional[bool] = False,
-                 client_id: Optional[str] = None):
+                 token_credential: Optional[TokenCredential] = None):
```
cgillum, line 17: Make this a required argument, moving it up right after the `taskhub` parameter.

durabletask-azuremanaged/durabletask/azuremanaged/client.py
```diff
@@ -27,14 +26,7 @@ def __init__(self, *,
...
         self._metadata.append(("dts", "True"))
-        self._metadata.append(("use_managed_identity", str(use_managed_identity)))
-        self._metadata.append(("client_id", str(client_id or "None")))
-
-        self._access_token_manager = AccessTokenManager(use_managed_identity=use_managed_identity,
-                                                        client_id=client_id)
-        token = self._access_token_manager.get_access_token()
-        self._metadata.append(("authorization", token))
-
+        self._metadata.append(("token_credential", token_credential))
```
cgillum, line 29: This should not be added to the `_metadata` collection. "Metadata" is for key-value pairs that get added to the request, which this is not. Instead, pass it to the `DTSDefaultClientInterceptorImpl` constructor directly. On top of that, there should be no need for `_metadata` at all in this class, so let's just remove it.

durabletask-azuremanaged/durabletask/azuremanaged/internal/access_token_manager.py
```diff
@@ -4,47 +4,43 @@
...
-                self._logger.debug("Using User Assigned Managed Identity for authentication.")
-                self.credential = ManagedIdentityCredential(client_id=self._client_id)
+        # Choose the appropriate credential. 
+        # Both TokenCredential and DefaultAzureCredential get_token methods return an AccessToken
+        if token_credential:
+            self._logger.debug("Using user provided token credentials.")
+            self._credential = token_credential
         else:
-            self.credential = DefaultAzureCredential()
+            self._credential = DefaultAzureCredential()
```
cgillum, line 22: Let's not specify any default credential. The user *must* provide their own credential. If they explicitly provide `None` as their credential, then we should skip all the token manager stuff entirely and make unauthenticated requests (which is valid in local testing setups).

examples/dts/dts_activity_sequence.py
```diff
@@ -47,7 +47,7 @@ def sequence(ctx: task.OrchestrationContext, _):
 
 
 # configure and start the worker
-with DurableTaskSchedulerWorker(host_address=endpoint, secure_channel=True, use_managed_identity=False, client_id="", taskhub=taskhub_name) as w:
+with DurableTaskSchedulerWorker(host_address=endpoint, secure_channel=True, taskhub=taskhub_name) as w:
```
cgillum, line 50: Per my other comment, the user should always provide the token credential, so let's update these samples and have them create and pass in their own `DefaultAzureCredential`.

cgillum (commented):
I think we're super close. A few more things to clean up.

durabletask-azuremanaged/durabletask/azuremanaged/durabletask_grpc_interceptor.py
```diff
@@ -3,28 +3,23 @@
...
 import grpc
 
 class DTSDefaultClientInterceptorImpl (DefaultClientInterceptorImpl):
     """The class implements a UnaryUnaryClientInterceptor, UnaryStreamClientInterceptor,
     StreamUnaryClientInterceptor and StreamStreamClientInterceptor from grpc to add an 
     interceptor to add additional headers to all calls as needed."""
 
-    def __init__(self, metadata: list[tuple[str, str]]):
+    def __init__(self, token_credential: TokenCredential, taskhub_name: str):
+        metadata = [("taskhub", taskhub_name)]
```
cgillum, line 15: Looks like this variable is assigned but not used? Did you mean to use `_metadata` here?

durabletask-azuremanaged/durabletask/azuremanaged/worker.py
```diff
@@ -12,22 +12,13 @@ class DurableTaskSchedulerWorker(TaskHubGrpcWorker):
     def __init__(self, *,
                  host_address: str,
                  taskhub: str,
-                 secure_channel: Optional[bool] = True,
-                 metadata: Optional[list[tuple[str, str]]] = None,
-                 token_credential: Optional[TokenCredential] = None):
+                 token_credential: TokenCredential = None,
```
cgillum, line 15: Let's remove the ` = None`. Let's make customers always be explicit about the `token_credential` parameter.

durabletask-azuremanaged/durabletask/azuremanaged/client.py
```diff
@@ -12,22 +12,13 @@ class DurableTaskSchedulerClient(TaskHubGrpcClient):
     def __init__(self, *,
                  host_address: str,
                  taskhub: str,
-                 secure_channel: Optional[bool] = True,
-                 metadata: Optional[list[tuple[str, str]]] = None,
-                 token_credential: Optional[TokenCredential] = None):
+                 token_credential: TokenCredential = None,
```
cgillum, line 15: Let's remove the ` = None`. Let's make customers always be explicit about the `token_credential` parameter.

cgillum (commented):
Looks like the tests are failing, but just a couple more things in addition to that.

durabletask/internal/shared.py
```diff
@@ -44,13 +45,14 @@ def get_grpc_channel(
...
 
+    # Create the base channel
     if secure_channel:
         channel = grpc.secure_channel(host_address, grpc.ssl_channel_credentials())
     else:
         channel = grpc.insecure_channel(host_address)
 
-    if metadata is not None and len(metadata) > 0:
-        interceptors = [DefaultClientInterceptorImpl(metadata)]
+    # Apply interceptors ONLY if they exist
```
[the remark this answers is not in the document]
cgillum: Nevermind, it looks like the calling code is already taking care of this. It looks like we can safely remove the `metadata` parameter from this function signature.

cgillum (commented):
OK, changes look good. I'll sign off after doing some local validation. I may add another commit as well depending on how things go.

cgillum (approved):

torosent (approved):
LGTM

berndverst:
I have a packaging concern:

Are we publishing `durabletask-azuremanaged` as a separate package? That file imports `azure-identity` and `azure-core` libraries. But those libraries are missing from `pyproject.toml`.

We really need to have a separate `pyproject.toml` to publish the `durabletask-azuremanaged` package separately and specify all required dependencies. Otherwise a user using that package will receive import errors because `azure-core` and `azure-identity` are missing - and that's not a good experience.

On the other hand, the main `durabletask` package isn't Azure specific - so for this reason we should not add these packages to the main `pyproject.toml` file.

durabletask-azuremanaged/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+classifiers = [
+  "Development Status :: 3 - Alpha",
+  "Programming Language :: Python :: 3",
+  "License :: OSI Approved :: MIT License",
+]
+requires-python = ">=3.9"
+license = {file = "LICENSE"}
+readme = "README.md"
+dependencies = [
+    "durabletask",
```
berndverst, line 29: Define a minimum required version here like `durabletask>=0.2.0` or ``durabletask>=0.2.1b` etc

durabletask-azuremanaged/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+    "azure-identity"
```
berndverst, line 30: Define a minimum required version -- likely you are not using something that was introduced recently so you can set a fairly low version, but nevertheless it's a good idea you have a minimum version. Otherwise customers / users may use existing incompatible versions or really old versions that do not have tools we might need for troubleshooting etc

requirements.txt
```diff
@@ -3,3 +3,5 @@ grpcio>=1.60.0 # 1.60.0 is the version introducing protobuf 1.25.X support, newe
 protobuf
 pytest
 pytest-cov
+azure-core
```
berndverst, line 6: Doesn't `azure-identity` import `azure-core` anyway? Maybe remove this unless you intent to set a minimum version.

Please see also my comment on the `pyproject` to specify a minimum version for `azure-identity`

examples/dts/README.md
```diff
@@ -0,0 +1,55 @@
...
+    export ENDPOINT=<taskhubEndpoint>
+    ```
+
+1. Since the samples rely on azure identity, ensure the package is installed and up-to-date
+
+    ```bash
+    python3 -m pip install azure-identity
+    ```
+
+1. Install the correct packages from the top level of this repository, i.e. durabletask-python/
```
berndverst, line 35: These are development instructions -- but wouldn't it be better to tell people to install from PyPI using `pip install durabletask` and `pip install durabletask-azuremanaged` ?

The instructions here would install and unreleased version from main branch -- we definitely will not provide support for installing from `main` branch. Only artifacts released to PyPI.

So I think making this clear would be worthwhile @RyanLettieri

examples/dts/dts_activity_sequence.py
```diff
@@ -0,0 +1,69 @@
...
+# Check if the variable exists
+if endpoint:
+    print(f"The value of ENDPOINT is: {endpoint}")
+else:
+    print("ENDPOINT is not set. Please set the ENDPOINT environment variable to the endpoint of the scheduler")
+    print("If you are using windows powershell, run the following: $env:ENDPOINT=\"<schedulerEndpoint>\"")
+    print("If you are using bash, run the following: export ENDPOINT=\"<schedulerEndpoint>\"")
+    exit()
+
+credential = DefaultAzureCredential()
```
berndverst, line 52: Do we want to add a comment that any `azure-identity` credential type and configuration can be used here instead?  For example, if someone wishes to use DefaultAzureCredential with a user managed identity this requires additional configuration here.

berndverst:
```
durabletask/azuremanaged/internal/access_token_manager.py:28: error: Incompatible types in assignment (expression has type "None", variable has type "datetime")  [assignment]
```

examples/dts/README.md
RyanLettieri: I'm not sure how up-to-date we keep these packages. I'll specify that this is installing from main as that is what the top level README.md does

berndverst: Customers should not be installing from main. Only contributors should be doing so - or folks who want to live on the edge and don't need support.

Also keep in mind that customers would never see any changes unless they checked out changes from this repo first. Pulling updates from PyPI is easier.

We will definitely want to publish this package frequently enough to PyPI - but keep in mind that we need to also support the packages we publish - which is easier to do than having to provide official support for every single commit in this repo.

Thoughts @torosent @cgillum ?

berndverst (approved):

examples/dts/README.md
cgillum: Yep, PyPI will be the only officially supported way to use our libraries.

PR #41, opened 2025-01-23
remove gitmodule file

YunchuWang:


YunchuWang (commented):
removing since we dont use submodule and this file can ensure git not init with submodule not needed

berndverst (approved):

PR #43, opened 2025-03-05
Creation of pipeline to publish dts python package to pypi

RyanLettieri:
This PR adds in a new workflow yaml that will be responsible for publishing the durabletask scheduler package to pypi. It also adds in e2e tests for DTS


.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Extract version from tag
+        run: echo "VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_ENV  # Extract version from the tag
```
berndverst, line 17:
```suggestion
        run: echo "VERSION=${GITHUB_REF#refs/tags/azuremanaged-v}" >> $GITHUB_ENV  # Extract version from the tag
```

.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,45 @@
+name: Publish Durable Task Scheduler to PyPI
+
+on:
+  push:
+    tags:
+      - 'v*'  # Only run for tags starting with "v"
```
berndverst, line 6:
```suggestion
      - 'azuremamanged-v*'  # Only run for tags starting with "azuremanaged-v"
```

.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  publish:
```
berndverst, line 9: before this intrdouce another job:

"test" which runs the unit tests, linter, type checker

Then make the "publish" job here depend on the "test" job via the `needs: {jobname}` directive.

berndverst (changes requested):
A few more changes required to ensure we produce correct releases.

.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Check package
+        run: |
+          cd durabletask-azuremanaged 
+          twine check dist/* 
+
+      - name: Publish package to PyPI
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}  # Store your PyPI API token in GitHub Secrets
```
berndverst, line 42:
```suggestion
          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN_AZUREMANAGED }}  # Store your PyPI API token in GitHub Secrets
```

.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"  # Adjust Python version as needed
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install build twine
+
+      - name: Build package from directory durabletask-azuremanaged
```
berndverst, line 29:
```suggestion
      - name: Build package from directory durabletask-azuremanaged
        working-directory: durabletask-azuremanaged
```

This block should look like:

```
      - name: Build package from directory durabletask-azuremanaged
        working-directory: ./durabletask-azuremanaged
        run: |
          python -m build
```

We don't need to `cd` which can get confusing. Let's just set the working directory (which is relative to the root)

.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: |
+          cd durabletask-azuremanaged 
+          python -m build 
+
+      - name: Check package
+        run: |
+          cd durabletask-azuremanaged
```
berndverst, line 36:
```suggestion
```

.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install build twine
+
+      - name: Build package from directory durabletask-azuremanaged
+        run: |
+          cd durabletask-azuremanaged 
+          python -m build 
+
+      - name: Check package
```
berndverst, line 34:
```suggestion
      - name: Check package
        working-directory: ./durabletask-azuremanaged
```

.github/workflows/publish-dts-sdk.yml
berndverst: FYI - right now we are not doing anything with the version. However, if we wanted to - we could use this version from the tag and DYNAMICALLY replace the value in pyproject.toml BEFORE running the build and publish commands. This way releasing a new SDK only requires tagging... otherwise what would be required is also a manual PR to change the version in pyproject -- and in fact what can happen is that the pyproject file version and the tag version are not aligned.

Consider having a script here that uses `sed` or similar to replace the version within the `pyproject.toml`

.github/workflows/pr-validation.yml
```diff
@@ -1,51 +1,58 @@
...
+        pip install flake8 pytest
+        pip install -r requirements.txt
+    - name: Install durabletask-azuremanaged dependencies
+      working-directory: examples/dts
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+    - name: Lint with flake8
+      run: |
+        flake8 . --count --show-source --statistics --exit-zero
```
berndverst, line 40:
```suggestion
        flake8 . --count --show-source --statistics

```

We want this step to fail if there are linter issues, so let's not use `--exit-zero`

.github/workflows/pr-validation.yml
```diff
@@ -1,51 +1,58 @@
...
+      uses: actions/setup-go@v5
+      with:
+        go-version: 'stable'
+
+    # Install and run the durabletask-go sidecar for running e2e tests
+    - name: Pytest e2e tests
+      working-directory: tests/durabletask
+      run: |
+        go install github.com/microsoft/durabletask-go@main
+        durabletask-go --port 4001 &
```
berndverst, line 57: doesn't this mean Python will be dependent on Go updates?

.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,111 @@
...
+
+  run-docker-tests:
+    needs: lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Pull Docker image
+        run: docker pull mcr.microsoft.com/dts/dts-emulator:v0.0.4
```
berndverst, line 43: Extract the tag version to a variable so we can reference it more easily and update it more easily in a single place.

examples/dts/requirements.txt
```diff
@@ -0,0 +1,8 @@
+autopep8
+grpcio>=1.60.0 # 1.60.0 is the version introducing protobuf 1.25.X support, newer versions are backwards compatible
+protobuf
+pytest
```
berndverst, line 4: `pytest` isn't a requirement for running the xample itself, and neither is `pytest-cov`

If you need to install this - move it to ` dev-requirements.txt` elsewhere?

.github/workflows/pr-validation.yml
berndverst: I see that was there before - but let's actually change this!

.github/workflows/pr-validation.yml
berndverst: EDIT: This was there before, so no need to address this - but something we want to consider.

berndverst (changes requested):
Left some comments - just some pipeline changes and dependency changes. Nothing major.

.github/workflows/pr-validation.yml
RyanLettieri: This leads to a lot of lint errors that were being ignored. I'll create an additional PR afterwards that removes this flag and addresses the linting problems so I don't clutter this PR too much

berndverst: Ok, please do that! Just because existing linter errors exist doesn't mean we should leave them in place.

.github/workflows/publish-dts-sdk.yml
```diff
@@ -0,0 +1,110 @@
...
+        python -m pip install --upgrade pip
+        pip install setuptools wheel tox
+        pip install flake8
+    - name: Run flake8 Linter
+      working-directory: durabletask-azuremanaged
+      run: flake8 .
+
+  run-docker-tests:
+    env:
+      EMULATOR_VERSION: "v0.0.4"  # Define the variable
```
berndverst, line 34:
```suggestion
      EMULATOR_VERSION: "v0.0.5"  # Define the variable
```

berndverst (approved):

PR #50, opened 2025-05-30
Fix and improve connection handling, add concurrency options, prep for release

berndverst:
Fix and improve connection handling, prep for release


durabletask/worker.py
```diff
@@ -129,31 +132,117 @@ def add_activity(self, fn: task.Activity) -> str:
...
+                    current_channel = shared.get_grpc_channel(self._host_address, self._secure_channel, self._interceptors)
+                    current_stub = stubs.TaskHubSidecarServiceStub(current_channel)
+
+                    # Test the connection
+                    current_stub.Hello(empty_pb2.Empty())
+                    conn_retry_count = 0  # Reset on successful connection
+                    self._logger.debug(f"Created fresh connection to {self._host_address}")
+
+                except Exception as e:
+                    self._logger.debug(f"Failed to create connection: {e}")
```
cgillum, line 176: Why only debug? Should the error level be higher?

durabletask/worker.py
```diff
@@ -91,13 +92,15 @@ def __init__(self, *,
                  log_handler=None,
                  log_formatter: Optional[logging.Formatter] = None,
                  secure_channel: bool = False,
-                 interceptors: Optional[Sequence[shared.ClientInterceptor]] = None):
+                 interceptors: Optional[Sequence[shared.ClientInterceptor]] = None,
+                 max_workers: Optional[int] = None):
```
cgillum, line 96: The .NET SDK allows configuring separate maximum values for orchestrator work items, activity work items, and entity work items (see [here](https://github.com/microsoft/durabletask-dotnet/blob/b2f3fed1e741da67620104aed0eec340bd1b0425/src/Worker/Core/DurableTaskWorkerOptions.cs#L123-L131)). We need to do the same for the Python SDK for consistency. These values also need to be provided to the DTS backend as part of the `GetWorkItems` request, which I noted below.

durabletask/worker.py
```diff
@@ -129,31 +132,117 @@ def add_activity(self, fn: task.Activity) -> str:
...
+                                break  # Shutdown requested during wait
+                            continue
 
-                        # stream work items
+                    try:
+                        # Stream work items with the current connection
+                        # Type assertion since we know current_stub is not None at this point
+                        assert current_stub is not None, "current_stub should not be None at this point"
+                        stub = current_stub  # Local reference for type safety
                         self._response_stream = stub.GetWorkItems(pb.GetWorkItemsRequest())
```
cgillum, line 234: As noted in my other comments, we should pass the user-provided concurrency values as parameters to `GetWorkItems`. This will ensure that DTS doesn't try to send more work items than can be handled by the worker.

durabletask/worker.py
```diff
@@ -11,20 +15,65 @@
 import grpc
 from google.protobuf import empty_pb2
 
-import durabletask.internal.helpers as ph
 import durabletask.internal.helpers as pbh
+import durabletask.internal.helpers as ph
```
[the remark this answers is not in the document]
berndverst: It was there before, but doesn't make sense. I'll clean it up.

PR #58, opened 2025-08-14
Add orchestration versioning support

andystaples:
Adds orchestration versioning support

Syntax: 

To add a versioning strategy to the worker: 
```python
    with DurableTaskSchedulerWorker(host_address=endpoint, secure_channel=True,
                                    taskhub=taskhub_name, token_credential=None) as w:
        w.add_orchestrator(single_activity)
        w.add_activity(plus_one)
        w.use_versioning(worker.VersioningOptions(
            version="1.0.0",
            default_version="1.0.0",
            match_strategy=worker.VersionMatchStrategy.CURRENT_OR_OLDER,
            failure_strategy=worker.VersionFailureStrategy.FAIL
        ))
        w.start()
```

To add a default version to the client (version to use when scheduling new orchestrations): 
```python
        task_hub_client = client.TaskHubGrpcClient(default_version="1.0.0")
```

To schedule an orchestration with a manual version: 
```python
        id = task_hub_client.schedule_new_orchestration(sequence, input=1, version="1.0.0")
```

To schedule a sub-orchestration with a manual version: 
```python
        yield ctx.call_sub_orchestrator(sequence_suborchestator, input=1, version='1.0.0')
```

### DIFF
--- a/durabletask/worker.py
+++ b/durabletask/worker.py
@@ -573,7 +573,7 @@ class TaskHubGrpcWorker:
                 customStatus=ph.get_string_value(result.encoded_custom_status),
                 completionToken=completionToken,
             )
-        except pe.AbandonOrchestrationError as ex:
+        except pe.AbandonOrchestrationError:
             self._logger.info(
                 f"Abandoning orchestration. InstanceId = '{req.instanceId}'. Completion token = '{completionToken}'"
             )
--- a/durabletask/worker.py
+++ b/durabletask/worker.py
@@ -651,6 +651,7 @@ class _RuntimeOrchestrationContext(task.OrchestrationContext):
         self._current_utc_datetime = datetime(1000, 1, 1)
         self._instance_id = instance_id
         self._registry = registry
+        self._version: Optional[str] = None
         self._completion_status: Optional[pb.OrchestrationStatus] = None
         self._received_events: dict[str, list[Any]] = {}
         self._pending_events: dict[str, list[task.CompletableTask]] = {}
@@ -775,6 +776,10 @@ class _RuntimeOrchestrationContext(task.OrchestrationContext):
     @property
     def instance_id(self) -> str:
         return self._instance_id
+
+    @property
+    def version(self) -> Optional[str]:
+        return self._version
 
     @property
     def current_utc_datetime(self) -> datetime:
@@ -977,11 +982,12 @@ class _OrchestrationExecutor:
 
             # Process versioning if applicable
             execution_started_events = [e.executionStarted for e in old_events if e.HasField("executionStarted")]
+            # We only check versioning if there are executionStarted events - otherwise, on the first replay when
+            # ctx.version will be Null, we may invalidate orchestrations early depending on the versioning strategy.
             if self._registry.versioning and len(execution_started_events) > 0:
-                execution_started_event = execution_started_events[-1]
                 version_failure = self.evaluate_orchestration_versioning(
                     self._registry.versioning,
-                    execution_started_event.version.value if execution_started_event.version else None,
+                    ctx.version
                 )
                 if version_failure:
                     self._logger.warning(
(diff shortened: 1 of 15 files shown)


durabletask/worker.py
```diff
@@ -279,6 +332,12 @@ def add_activity(self, fn: task.Activity) -> str:
             )
         return self._registry.add_activity(fn)
 
+    def use_versioning(self, version: VersioningOptions) -> None:
+        """Sets the default version for orchestrators and activities."""
```
andystaples, line 336: Need to improve this comment

durabletask/worker.py
```diff
@@ -72,9 +74,60 @@ def __init__(
...
+
+    def __init__(self, version: Optional[str] = None,
+                 default_version: Optional[str] = None,
+                 match_strategy: Optional[VersionMatchStrategy] = None,
+                 failure_strategy: Optional[VersionFailureStrategy] = None
+                 ):
+        """Initialize versioning options.
+
+        Args:
+            version: The specific version to use for orchestrators and activities.
```
andystaples, line 116: Need to update these comments
version = the "worker version" that will be used for comparison
default_version = the version that will be assigned to new sub-orchestrations started by this worker

durabletask/worker.py
```diff
@@ -778,6 +838,7 @@ def call_sub_orchestrator(
             retry_policy=retry_policy,
             is_sub_orch=True,
             instance_id=instance_id,
+            version=version,
```
andystaples, line 841: Need to find out how to get the default version from worker and use it here if version from the params is not specified

andystaples (commented):
Notes to self:

tests/durabletask-azuremanaged/test_dts_orchestration_versioning_e2e.py
```diff
@@ -0,0 +1,319 @@
...
+import json
+import os
+
+import pytest
+
+from durabletask import client, task, worker
+from durabletask.azuremanaged.client import DurableTaskSchedulerClient
+from durabletask.azuremanaged.worker import DurableTaskSchedulerWorker
+
+# NOTE: These tests assume a sidecar process is running. Example command:
```
berndverst, line 13: This comment probably doesn't make sense - but maybe upate it with instructions to spin up the DTS emulator and run against that instead?

andystaples: Done - updated all the e2e instructions, including the durabletask ones pointing to durabletask-go instead of chris's sidecar fork

berndverst (commented):
The PR looks good. The one thing I am wondering about:
Am I able to actually provide two versions of my orchestration that live side by side (say 1.2.0 and 1.2.2) and route to the correct version according to either the version I requested or my policies?
I don't see any tests covering this and it's not clear to me whether / how this is possible.

andystaples:
> The PR looks good. The one thing I am wondering about: Am I able to actually provide two versions of my orchestration that live side by side (say 1.2.0 and 1.2.2) and route to the correct version according to either the version I requested or my policies? I don't see any tests covering this and it's not clear to me whether / how this is possible.

Short answer... not really
Long answer: The logic that picks up items from the task hub and distributes them to Durable workers for processing isn't version-aware. It simply finds the next available worker and assigns it the oldest item that still needs processing from the task hub. So, if you have two orchestrations, 1.2.0 and 1.2.2, and two workers with the corresponding versions, it is pseudo-random which worker will process what. That's why we offer VersionFailureStrategy.REJECT, so that if a work item goes to a worker not capable of processing it, you can keep reprocessing it until hopefully it gets assigned to a valid worker
The feature is definitely not designed to have multiple versions of the same app running at the same time for long periods of time, though. You'd want separate task hubs for that. It's more for the short-term during high availability deployments

PR #63, opened 2025-09-26
Add Durable Entities support

andystaples:
Add support for durable entities to this SDK

See the included documentation/samples/tests for usage guidelines. 

Reviewers: Request special attention to file structures, naming, docs and tests

Discussion points: 
* Optional parameters to some of the entity requests, like start time and cancellation tokens - should these be implemented now or later?
* Typing for entity calls - look at TInput/TOutput for activities, compare with entities (abstraction because of EntityInstanceId) 
* Class-based entities - should we have the object state represent the entity state, or preserve the get_state & from_state behavior here (research into the dotnet SDK required)

### DIFF
--- a/durabletask/entities/entity_lock.py
+++ b/durabletask/entities/entity_lock.py
@@ -1,14 +1,19 @@
+import durabletask.internal.helpers as ph
+
 from durabletask.entities.entity_instance_id import EntityInstanceId
-from durabletask.internal.orchestration_entity_context import OrchestrationEntityContext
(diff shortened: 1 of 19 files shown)


durabletask/entities/entity_lock.py
```diff
@@ -0,0 +1,17 @@
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from durabletask.task import OrchestrationContext
+
+
+class EntityLock:
+    # Note: This should
```
andystaples, line 9: Will clean this up

PR #75, opened 2025-11-21
Andystaples/add functions support

andystaples:
DRAFT PR - for collaboration
Requires at minimum the changes here 
https://github.com/Azure/azure-functions-durable-extension/pull/3260
in durable WebJobs extension allowing gRPC protocol for Python


durabletask-azurefunctions/durabletask/azurefunctions/decorators/durable_app.py
```diff
@@ -0,0 +1,266 @@
...
+                stub.CompleteOrchestratorTask = stub_complete
+                execution_started_events = [e for e in [e1 for e1 in request.newEvents] + [e2 for e2 in request.pastEvents] if e.HasField("executionStarted")]
+                function_name = execution_started_events[-1].executionStarted.name
+                worker.add_named_orchestrator(function_name, orchestrator_func)
+                worker._execute_orchestrator(request, stub, None)
+
+                if response is None:
+                    raise Exception("Orchestrator execution did not produce a response.")
+                # The Python worker returns the input as type "json", so double-encoding is necessary
+                return '"' + base64.b64encode(response.SerializeToString()).decode('utf-8') + '"'
```
andystaples: Victoria - Currently, the return value from here is passed on to the host as type "json" so the host attempts to Newtonsoft deserialize it back into an object before handing back to the Durable middleware for final decoding. This breaks, unless I double-encode with quotes as above. Is there a way to communicate to the worker that this is a plain string instead?

durabletask-azurefunctions/durabletask/azurefunctions/decorators/durable_app.py
```diff
@@ -0,0 +1,266 @@
...
+        return decorator
+
+    def _add_rich_client(self, fb, parameter_name,
+                         client_constructor):
+        # Obtain user-code and force type annotation on the client-binding parameter to be `str`.
+        # This ensures a passing type-check of that specific parameter,
+        # circumventing a limitation of the worker in type-checking rich DF Client objects.
+        # TODO: Once rich-binding type checking is possible, remove the annotation change.
+        user_code = fb._function._func
+        user_code.__annotations__[parameter_name] = str
```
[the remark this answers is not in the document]
andystaples: We'd have to figure out how to detect which underlying provider for the durable_client_input binding is being used to know when to simply return the string for the old SDK vs parse it in the new

durabletask-azurefunctions/pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+
+# For more information on pyproject.toml, see https://peps.python.org/pep-0621/
+
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "durabletask.azurefunctions"
+version = "0.1.0"
```
andystaples: Victoria - what versioning strategy would you propose if the first version that goes to PyPi would be used for internal testing only?

durabletask-azurefunctions/durabletask/azurefunctions/decorators/durable_app.py
```diff
@@ -0,0 +1,266 @@
...
+                request = OrchestratorRequest()
+                request.ParseFromString(base64.b64decode(orchestration_context))
+                stub = AzureFunctionsNullStub()
+                worker = DurableFunctionsWorker()
+                response: Optional[OrchestratorResponse] = None
+
+                def stub_complete(stub_response):
+                    nonlocal response
+                    response = stub_response
+                stub.CompleteOrchestratorTask = stub_complete
```
andystaples: All of this is probably optimizable - do we really need to create a new stub and worker for each call? Can they be saved? Will look into this more at some point

durabletask-azurefunctions/durabletask/azurefunctions/decorators/durable_app.py
hallvictoria: This is actually coming from the OrchestrationTriggerConverter done [here](https://github.com/Azure/azure-functions-python-library/blob/8bd30d300529cbda0526aa2d6606d11333c9c3aa/azure/functions/durable_functions.py#L32). The type is hard-coded to json. Changing the type to `string` works in this scenario, but I'm unsure if there are other cases where `json` was intended

```
@classmethod
    def encode(cls, obj: typing.Any, *, expected_type: typing.Optional[type]) -> meta.Datum:
        # Durable function context should be a json
        return meta.Datum(type='json', value=obj)
```

PR #113, opened 2026-02-19
Add In-Process backend for E2E testing

andystaples:
Adds a fully-featured in-memory Durable backend for E2E testing. Removes the previous dependency on the durabletask-go sidecar. Implement missing tests for `durabletask` including tests for batch actions, entities, and versioning.

### DIFF
--- a/.github/workflows/durabletask-azuremanaged-dev.yml
+++ b/.github/workflows/durabletask-azuremanaged-dev.yml
@@ -8,6 +8,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   publish-dev:
     runs-on: ubuntu-latest
--- a/.github/workflows/durabletask-azuremanaged-experimental.yml
+++ b/.github/workflows/durabletask-azuremanaged-experimental.yml
@@ -6,6 +6,9 @@ on:
       - main
       - release/*
 
+permissions:
+  contents: read
+
 jobs:
   publish-experimental:
     runs-on: ubuntu-latest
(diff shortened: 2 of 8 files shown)


durabletask/testing/in_memory_backend.py
```diff
@@ -0,0 +1,1556 @@
...
+            for e in self._entities.values():
+                if str(e.completion_token) == request.completionToken:
+                    entity = e
+                    break
+
+            if not entity:
+                self._logger.warning(
+                    f"No entity found for completion token '{request.completionToken}'"
+                )
+                return pb.CompleteTaskResponse()
```
halspang: Does this state signify anything and would we need to potentially clean up something? Or can we not know since the completion token isn't valid anymore?

andystaples: We won't be able to clean anything up, since the completion token is no longer valid. I imagine this would only be the case if CompleteEntityTask got called multiple times for the same token, in which case short-circuiting to return early is the correct behavior anyway.

durabletask/testing/in_memory_backend.py
```diff
@@ -0,0 +1,1556 @@
...
+            if instance and predicate(instance):
+                return instance
+
+            waiter = StateWaiter(predicate=predicate)
+            if instance_id not in self._state_waiters:
+                self._state_waiters[instance_id] = []
+            self._state_waiters[instance_id].append(waiter)
+
+        # Wait outside the lock
+        wait_result = waiter.event.wait(timeout=timeout if timeout else 30.0)
```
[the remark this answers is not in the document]
andystaples: Are there other wait methods? The only wait method we define is this _wait_for_state() with the 30 second default. The other places we wait in this backend are checking for new work items with a 0.1s wait equivalent to a non-blocking read, and during shutdown

PR #121, opened 2026-03-06
Add rewind support

### LINKED ISSUE #68: Add support for orchestration rewind
See https://github.com/Azure/azure-functions-durable-extension/pull/3227

andystaples:
Adds rewind support to `durabletask`. 
Requires the following changes: 
1. Add rewind action to protobuf: https://github.com/microsoft/durabletask-protobuf/pull/65
2. Add rewind support to the DTS sidecar: https://msazure.visualstudio.com/One/_git/AAPT-DTMB/pullrequest/14935863
3. Add rewind action support to DTS

Resolves #68


durabletask/client.py
```diff
@@ -782,6 +785,26 @@ def resume_orchestration(self, instance_id: str) -> None:
         self._logger.info(f"Resuming instance '{instance_id}'.")
         self._stub.ResumeInstance(req)
 
+    def rewind_orchestration(self, instance_id: str, *,
```
berndverst, line 791: 🔴 **Async client parity.** This adds `rewind_orchestration` only to the sync `TaskHubGrpcClient`. `AsyncTaskHubGrpcClient` further down this file has async `terminate`/`suspend`/`resume`/`restart`/`purge` but no `rewind_orchestration`, so `AsyncDurableTaskSchedulerClient` (durabletask-azuremanaged) inherits nothing and can't rewind.

Please add an `async def rewind_orchestration(...)` mirroring this method (with `await self._stub.RewindInstance(req)`), plus the matching declaration on `_AsyncTaskHubSidecarServiceStub` (see the sync stub addition above at `RewindInstance`). The generated `orchestrator_service_pb2_grpc.py` already supports the RPC, so this is just wiring. Ideally add an async DTS rewind test as well.

durabletask/client.py
```diff
@@ -320,6 +320,9 @@ def SuspendInstance(self, request: pb.SuspendRequest) -> pb.SuspendResponse:
     def ResumeInstance(self, request: pb.ResumeRequest) -> pb.ResumeResponse:
         ...
 
+    def RewindInstance(self, request: pb.RewindInstanceRequest) -> pb.RewindInstanceResponse:
```
berndverst, line 323: Mirror this on `_AsyncTaskHubSidecarServiceStub` too — it currently has no `async def RewindInstance(...)` declaration, which is part of why the async client can't call the RPC under type checking.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+ADDED
+
```
berndverst, line 11: Rewind is also surfaced through `DurableTaskSchedulerClient` / `AsyncDurableTaskSchedulerClient` and has a dedicated `tests/durabletask-azuremanaged/test_dts_rewind_e2e.py`. Per this repo's guideline ("if a change affects both packages, update both changelogs"), please add a matching `## Unreleased` entry to `durabletask-azuremanaged/CHANGELOG.md`.

durabletask/worker.py
```diff
@@ -2206,6 +2226,97 @@ def execute(
...
+        failed_task_ids: set[int] = set()
+        for event in all_events:
+            if event.HasField("taskFailed"):
+                failed_task_ids.add(event.taskFailed.taskScheduledId)
+            elif event.HasField("subOrchestrationInstanceFailed"):
+                failed_task_ids.add(event.subOrchestrationInstanceFailed.taskScheduledId)
+
+        # Second pass: build the clean history.
+        clean_history: list[pb.HistoryEvent] = []
+        for event in all_events:
```
berndverst, line 2296: 🟡 **Retry-policy / timer interaction.** The second pass removes `taskFailed` and the failed `taskScheduled` events, but any `timerCreated`/`timerFired` events emitted between retry attempts (for an activity scheduled with a `RetryPolicy`) are kept. On the next replay the regenerated `scheduleTask` action won't line up with those orphaned timer events, which could trigger a non-determinism mismatch. Is a rewound activity that had a retry policy covered by a test? If this is a known/shared limitation with the Core rewind, it'd be worth calling out explicitly in the docstring.

durabletask/worker.py
```diff
@@ -2206,6 +2226,97 @@ def execute(
...
+
+        # First pass: collect the task-scheduled IDs that correspond to
+        # failed activities / sub-orchestrations so we can remove the
+        # matching taskScheduled events in the second pass.
+        failed_task_ids: set[int] = set()
+        for event in all_events:
+            if event.HasField("taskFailed"):
+                failed_task_ids.add(event.taskFailed.taskScheduledId)
+            elif event.HasField("subOrchestrationInstanceFailed"):
+                failed_task_ids.add(event.subOrchestrationInstanceFailed.taskScheduledId)
```
berndverst: Minor: this adds `subOrchestrationInstanceFailed.taskScheduledId` to `failed_task_ids`, but the second pass only uses that set to drop `taskScheduled` events by `eventId`. Failed sub-orchestrations are represented by `subOrchestrationInstanceCreated` (kept intentionally) and `subOrchestrationInstanceFailed` (dropped by its own check), so this ID never matches a `taskScheduled` event — this `elif` branch is effectively dead. The test docstring's claim that "their corresponding taskScheduled events are removed" reflects the same confusion. Suggest either removing this branch or adding a comment clarifying the intent.

berndverst (commented):
## Rewind support review (compared against durabletask-dotnet)

Thanks for this — overall it looks solid, and I appreciate that the SDK-side history rewriting in `_build_rewind_result` is clearly documented with the backend-contract warning. I compared it against dotnet's rewind support and have a few gaps/notes (most are inline).

**🔴 Async client parity is missing (main gap).** `rewind_orchestration` and the stub `RewindInstance` are added only to the sync `TaskHubGrpcClient` / `_SyncTaskHubSidecarServiceStub`. `AsyncTaskHubGrpcClient` and `_AsyncTaskHubSidecarServiceStub` don't get equivalents, so `AsyncDurableTaskSchedulerClient` in durabletask-azuremanaged can't rewind. Every other lifecycle op (terminate/suspend/resume/restart/purge) exists on both clients, and the generated `orchestrator_service_pb2_grpc.py` already exposes the RPC. See inline.

**🟡 azuremanaged CHANGELOG.** Only the core `CHANGELOG.md` was updated, but rewind is surfaced through the DTS clients and has a dedicated `tests/durabletask-azuremanaged/test_dts_rewind_e2e.py`. Per the repo's dual-changelog guideline, `durabletask-azuremanaged/CHANGELOG.md` should get an `## Unreleased` entry. See inline.

**🟡 Retry-policy / timer edge case** in `_build_rewind_result` — see inline.

**🟢 Minor:** likely-dead branch around `subOrchestrationInstanceFailed.taskScheduledId` — see inline.

**🟢 Error mapping (not blocking).** dotnet maps gRPC status codes to typed exceptions (NotFound→ArgumentException, FailedPrecondition→InvalidOperationException, Unimplemented→NotImplementedException, Cancelled→OperationCanceledException). This PR lets `grpc.RpcError` propagate, which is consistent with every other Python client method here, so it's fine to leave — just flagging the UX difference from dotnet.

**Architecture note (no action needed).** dotnet keeps rewind a thin client→sidecar call and lets DurableTask.Core rewrite history (its worker only deserializes `ExecutionRewound`). This PR instead computes the clean history in the SDK worker. That's a reasonable divergence given the DTS backend design, but it creates a tight SDK↔backend contract: the worker detects a rewind via an `executionCompleted` event in committed history plus an `executionRewound` event in the new events. Since the DTS E2E tests are `@skip` until the emulator ships rewind, that contract is currently only exercised against the in-memory backend — worth a follow-up to un-skip once the emulator supports it.

durabletask/testing/in_memory_backend.py
```diff
@@ -1822,6 +1876,128 @@ def fire() -> None:
...
+        instance.dispatched_events.clear()
+
+        # Add the ExecutionRewound event as a new pending event.
+        rewind_event = pb.HistoryEvent(
+            eventId=-1,
+            timestamp=timestamp_pb2.Timestamp(),
+            executionRewound=pb.ExecutionRewoundEvent(
+                reason=wrappers_pb2.StringValue(value=reason) if reason else None,
+            ),
+        )
```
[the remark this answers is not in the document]
andystaples: Fixed. The `ExecutionRewound` event is now stamped with `Timestamp.FromDatetime(datetime.now(timezone.utc))` instead of the default epoch, consistent with other backend-generated events.

durabletask/testing/in_memory_backend.py
```diff
@@ -1822,6 +1876,128 @@ def fire() -> None:
...
+                rewind_event = event
+                break
+        instance.pending_events.clear()
+        instance.dispatched_events.clear()
+        if rewind_event is not None:
+            instance.pending_events.append(rewind_event)
+        instance.completion_token = self._next_completion_token
+        self._next_completion_token += 1
+        self._orchestration_in_flight.discard(instance.instance_id)
+        self._enqueue_orchestration(instance.instance_id)
```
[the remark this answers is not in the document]
andystaples: Fixed. `_process_rewind_orchestration_action` no longer re-appends `executionRewound` to `pending_events` — it's already present in the SDK-rewritten clean history. The post-rewind replay is now triggered with a lone `orchestratorStarted`, so `executionRewound` is no longer duplicated in the instance history when the dispatched events are committed.

durabletask/client.py
```diff
@@ -782,6 +785,26 @@ def resume_orchestration(self, instance_id: str) -> None:
         self._logger.info(f"Resuming instance '{instance_id}'.")
         self._stub.ResumeInstance(req)
 
+    def rewind_orchestration(self, instance_id: str, *,
+                             reason: str | None = None):
```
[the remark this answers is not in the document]
andystaples: Fixed. Added an explicit `-> None` return annotation to `rewind_orchestration` on both the sync `TaskHubGrpcClient` and the new async `AsyncTaskHubGrpcClient`.

tests/durabletask/test_rewind_e2e.py
```diff
@@ -0,0 +1,354 @@
...
+from durabletask import client, task, worker
+from durabletask.testing import create_test_backend
+
+HOST = "localhost:50055"
+
+
+@pytest.fixture(autouse=True)
+def backend():
+    """Create an in-memory backend for testing."""
+    b = create_test_backend(port=50055)
```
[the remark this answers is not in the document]
andystaples: Fixed. The test now uses a module-level `PORT = find_free_port()` (from `tests.durabletask._port_utils`) for both `HOST` and `create_test_backend(port=PORT)`, matching the other E2E tests.

durabletask/worker.py
```diff
@@ -2206,6 +2226,97 @@ def execute(
...
+            # parent's execution ID so it matches the parent's new run.
+            if event.HasField("executionStarted"):
+                event_copy = pb.HistoryEvent()
+                event_copy.CopyFrom(event)
+                event_copy.executionStarted.orchestrationInstance.executionId.CopyFrom(
+                    ph.get_string_value_or_empty(new_execution_id))
+                if rewind_event.HasField("parentExecutionId"):
+                    if rewind_event.parentExecutionId.value:
+                        event_copy.executionStarted.parentInstance.orchestrationInstance.executionId.CopyFrom(
+                            rewind_event.parentExecutionId)
```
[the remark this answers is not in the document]
andystaples: Fixed. The copy now only writes `parentInstance.orchestrationInstance.executionId` when the original `executionStarted` actually has a `parentInstance` (guarded with `HasField("parentInstance")` in addition to the `parentExecutionId` checks), so a top-level orchestration won't get an empty `parentInstance` materialized.

durabletask/client.py
andystaples: Done. Added `async def RewindInstance(self, request: pb.RewindInstanceRequest) -> pb.RewindInstanceResponse` to `_AsyncTaskHubSidecarServiceStub`, mirroring the sync stub addition, so the async client can call the RPC under type checking.

durabletask/worker.py
```diff
@@ -2140,6 +2140,26 @@ def execute(
                 "The new history event list must have at least one event in it."
             )
 
+        # Check for rewind BEFORE replay.  A rewind is indicated by an
+        # executionRewound event in new_events.  We look for an
+        # executionCompleted event anywhere in the history (old or new
+        # events) to decide whether to rewind or replay:
```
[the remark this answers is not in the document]
andystaples: Fixed. The comment now matches the code — it describes checking `executionCompleted` in `old_events` (the committed history), which is what the condition actually does.

durabletask/client.py
andystaples: Done. Added `async def rewind_orchestration(...)` to `AsyncTaskHubGrpcClient` mirroring the sync method (with `await self._stub.RewindInstance(req)`), so `AsyncDurableTaskSchedulerClient` now inherits it. The async stub declaration was added too (see the reply on the stub thread). The DTS rewind e2e suite is now enabled — the emulator supports `RewindInstance` as of the current `latest` image.

durabletask/worker.py
andystaples: Good catch — this is a known limitation shared with the Core/.NET rewind. I've called it out explicitly in the `_build_rewind_result` docstring: timer events (`timerCreated`/`timerFired`) emitted between retry attempts of an activity scheduled with a `RetryPolicy` are not removed, so the regenerated `scheduleTask` may not line up with those retained timer events on replay — meaning rewinding an activity that used a retry policy is not currently supported. There is no test for a retry-policy rewind precisely because it isn't supported.

CHANGELOG.md
andystaples: Done. Added a matching `## Unreleased` entry to `durabletask-azuremanaged/CHANGELOG.md` noting that `DurableTaskSchedulerClient` / `AsyncDurableTaskSchedulerClient` now surface `rewind_orchestration()`.

durabletask/worker.py
andystaples: Fixed. Removed the dead `elif subOrchestrationInstanceFailed` branch from the first pass — `failed_task_ids` now only collects `taskFailed.taskScheduledId` (the only IDs that match `taskScheduled` events dropped in the second pass). Failed sub-orchestrations remain handled by keeping `subOrchestrationInstanceCreated` and dropping `subOrchestrationInstanceFailed`. Also corrected the misleading test docstring that claimed sub-orchestrations' `taskScheduled` events are removed.

durabletask/worker.py
```diff
@@ -2140,6 +2140,26 @@ def execute(
...
+        # Check for rewind BEFORE replay.  A rewind is indicated by an
+        # executionRewound event in new_events.  We look for an
+        # executionCompleted event in the committed history (old_events)
+        # to decide whether to rewind or replay:
+        # 1. executionCompleted IS present → the orchestration reached a
+        #    terminal state (e.g. failed).  This is a *new* rewind that
+        #    the worker must short-circuit by building clean history.
+        # 2. executionCompleted is NOT present → the backend already
+        #    processed the RewindOrchestrationAction which removed
+        #    executionCompleted from history.  This is a normal
```
sophiatev: Super nit: I think this comment is confusing because it leaves the reader wondering why an `executionRewound` event is sent as part of the new events for a "normal replay". I would change this to instead say that this event is used as a "jump-start" to trigger the rewound Activities to rerun

andystaples: Good point. Reworded case 2 of that comment to explain that the `executionRewound` event here acts as a "jump-start": it wakes the orchestration so that normal replay re-emits `scheduleTask` actions for the removed activities, causing the previously-failed work to rerun (no further history rewrite needed). That should remove the "why is a rewound event part of a normal replay?" confusion.

berndverst (approved):

PR #135, opened 2026-04-24
Add gRPC client and worker connection resiliency

berndverst:
## Summary
- Add public gRPC resiliency option types and shared transport helpers, then wire them through core and Azure Managed constructors.
- Harden worker, sync client, and async client connection recovery to better survive silent disconnects, transport failures, and channel recreation scenarios.
- Add focused regression coverage plus user-facing changelog and design/plan updates for the new resiliency behavior.

## Test Plan
- [x] Focused pytest run covering grpc resiliency, worker resiliency, worker concurrency loop, client resiliency, and Azure Managed wrapper wiring.
- [x] flake8 on all changed source and test files.


tests/durabletask/test_client.py
```diff
@@ -290,6 +315,506 @@ def test_async_client_uses_provided_channel_directly():
...
+            rpc_task = asyncio.create_task(client.get_orchestration_state("abc"))
+            while first_stub.GetInstance.await_count == 0:
+                await asyncio.sleep(0)
+            close_task = asyncio.create_task(client.close())
+            await asyncio.sleep(0)
+        finally:
+            client._recreate_lock.release()
+
+        with pytest.raises(grpc.aio.AioRpcError):
+            await rpc_task
```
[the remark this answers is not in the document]
berndverst: Fixed. The rpc-task await in this test is now explicitly assigned, which keeps the assertion path intact and avoids the no-effect warning.

tests/durabletask/test_client.py
```diff
@@ -290,6 +315,506 @@ def test_async_client_uses_provided_channel_directly():
...
+        await close_task
```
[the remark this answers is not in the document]
berndverst: Fixed. The close-task await is now explicitly assigned as well, so the cleanup step is still exercised without leaving a no-effect expression statement.

durabletask/client.py
```diff
@@ -531,6 +680,64 @@ async def __aenter__(self):
...
+            close_task.add_done_callback(self._retired_channel_close_tasks.discard)
+
+    async def _close_retired_channel(self, channel: grpc.aio.Channel) -> None:
+        try:
+            await asyncio.sleep(30.0)
+            await channel.close()
+        finally:
+            try:
+                self._retired_channels.remove(channel)
+            except ValueError:
```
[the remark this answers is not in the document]
berndverst: Fixed. `_close_retired_channel` now removes the channel only if it is still present under `_recreate_lock`, which makes the cleanup idempotent without relying on an empty `except ValueError`.

durabletask/worker.py
```diff
@@ -115,6 +123,80 @@ class VersionFailureStrategy(Enum):
...
+                state.close_when_released = True
+
+        if channel_to_close is not None:
+            self._close_channel(channel_to_close)
+
+    @staticmethod
+    def _close_channel(channel: Any) -> None:
+        try:
+            channel.close()
+        except Exception:
```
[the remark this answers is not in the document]
berndverst: Fixed. `_InFlightChannelTracker._close_channel` still treats cleanup as best-effort, but it now logs suppressed close failures at debug level with stack traces.

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,49 @@
...
+import grpc
+
+LONG_POLL_METHODS = {"WaitForInstanceStart", "WaitForInstanceCompletion"}
+
+
+def get_full_jitter_delay_seconds(
+        attempt: int,
+        *,
+        base_seconds: float,
+        cap_seconds: float) -> float:
```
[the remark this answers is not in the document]
berndverst: Fixed. Reflowed the helper signature to standard indentation.

docs/superpowers/plans/2026-04-23-grpc-resiliency.md
```diff
@@ -0,0 +1,852 @@
+# gRPC Resiliency Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
```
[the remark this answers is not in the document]
berndverst: Fixed. Converted the plan note to a GitHub-style callout block.

durabletask/worker.py
```diff
@@ -590,44 +703,89 @@ def run_loop():
...
-                current_channel = self._channel if self._channel is not None else None
                 current_stub = None
                 raise
 
-        def invalidate_connection():
+        def wrap_execution(handler, release):
+            def wrapped(*args, **kwargs):
+                result = handler(*args, **kwargs)
+                release()
+                return result
```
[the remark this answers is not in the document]
berndverst: Fixed. `wrap_execution` now releases the in-flight channel reference from a `finally`, and I added a regression test that exercises a handler failure.

durabletask/client.py
```diff
@@ -166,24 +176,109 @@ def __init__(self, *,
...
         self._logger = shared.get_logger("client", log_handler, log_formatter)
         self.default_version = default_version
         self._payload_store = payload_store
 
+    def _invoke_unary(
+            self,
+            method_name: str,
+            request: Any,
+            *,
+            timeout: Optional[int] = None):
```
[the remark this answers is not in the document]
berndverst: Fixed. The unary helper and the public wait APIs now accept float timeouts so the annotations match gRPC's timeout contract.

durabletask/client.py
```diff
@@ -531,6 +680,64 @@ async def __aenter__(self):
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         await self.close()
 
+    async def _invoke_unary(
+            self,
+            method_name: str,
+            request: Any,
+            *,
+            timeout: Optional[int] = None):
```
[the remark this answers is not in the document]
berndverst: Fixed. The async unary helper and async wait APIs now accept float timeouts as well.

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[1/10] Centralize resiliency in a `UnaryUnaryClientInterceptor` instead of `_invoke_unary`**

The PR rewrites ~38 stub call sites in `durabletask/client.py` (sync + async) to go through `self._invoke_unary("MethodName", req)`. That has two ongoing-maintenance costs:

- The `"MethodName"` string keys lose static type help; a future rename of a stub method silently bypasses resiliency.
- Any new RPC added later must remember to be wrapped, or it won't participate in failure tracking / channel recreation.

The dotnet counterpart (PR #708) solves this with a single `ChannelRecreatingCallInvoker` that wraps the channel once. The Python equivalent is `grpc.intercept_channel` + a `UnaryUnaryClientInterceptor` (and `grpc.aio.UnaryUnaryClientInterceptor` for the async client). Doing it that way would let you:

- Delete `_invoke_unary` and revert all 38 call-site edits.
- Keep `LONG_POLL_METHODS` keyed on the gRPC method path (e.g. `/TaskHubSidecarService/WaitForInstanceStart`), matching the dotnet implementation.
- Have exactly one place to maintain for any future stub additions.

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[2/10] Align naming for the "SDK-owned channel" concept**

`durabletask/client.py` already exposes `self._owns_channel` to mean "the SDK created the channel."

`durabletask/worker.py` introduces a new helper for the same concept but inverts the framing:

```python
def _can_recreate_channel(self) -> bool:
    return self._channel is None
```

Here `self._channel` is the *caller-provided* channel, so `_can_recreate_channel()` actually tests "no caller-provided channel" — the same thing `_owns_channel` already names in the client. The opposite-sounding names for the same idea make the cross-file code harder to read.

Suggestion: pick one name (e.g., `_uses_sdk_channel` or just `_owns_channel`) and use it consistently in both `client.py` and `worker.py`.

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[3/10] `FailureTracker` mutates state without a lock; sync client is multi-threaded**

`durabletask/internal/grpc_resiliency.py`:

```python
@dataclass
class FailureTracker:
    threshold: int
    consecutive_failures: int = 0

    def record_failure(self) -> bool:
        ...
        self.consecutive_failures += 1
        return self.consecutive_failures >= self.threshold

    def record_success(self) -> None:
        self.consecutive_failures = 0
```

The sync `TaskHubGrpcClient` is commonly called from multiple worker threads (gRPC sync clients are typically used that way), and the read-modify-write `self.consecutive_failures += 1` is not atomic. The CPython GIL makes a corrupted counter unlikely in practice, but races can:

- Cause two threads to simultaneously cross the threshold and double-recreate (the cooldown mostly hides this).
- Cause `record_success` to lose a concurrent failure increment, masking a real transport problem.

Suggestion: add a `threading.Lock` inside `FailureTracker` (the async client only ever touches it from the single event loop, so an `asyncio.Lock` is unnecessary for that side). Alternatively split sync/async tracker variants.

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[4/10] Don't depend on the private `ThreadPoolExecutor._shutdown` attribute**

`durabletask/worker.py` (in `_AsyncWorkerManager`) reads CPython-private state in four places:

```python
if getattr(self.thread_pool, "_shutdown", False):
    self.thread_pool = self._create_thread_pool()
...
if not getattr(self.thread_pool, "_shutdown", False):
    self.thread_pool.shutdown(wait=True)
```

`_shutdown` is implementation-private — it isn't part of `concurrent.futures.ThreadPoolExecutor`'s public API and could be renamed or removed in a future Python release without notice.

Suggestion: track an explicit `self._pool_is_shutdown: bool` flag set inside `shutdown()` / `prepare_for_run()` and read everywhere instead of `getattr(..., "_shutdown", False)`.

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[5/10] `wrap_execution` and `wrap_cancellation` in `_async_run_loop` are identical**

In `durabletask/worker.py._async_run_loop`:

```python
def wrap_execution(handler, release):
    def wrapped(*args, **kwargs):
        try:
            return handler(*args, **kwargs)
        finally:
            release()
    return wrapped

def wrap_cancellation(handler, release):
    def wrapped(*args, **kwargs):
        try:
            return handler(*args, **kwargs)
        finally:
            release()
    return wrapped
```

These are byte-for-byte the same function with different names. Collapse to one helper (e.g. `_with_release(handler, release)`) and use it for both the execution and cancellation paths in `submit_work_item`.

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[6/10] `_async_run_loop` is hard to navigate; consider extracting a `_WorkItemStreamConsumer`**

After this PR, `TaskHubGrpcWorker._async_run_loop` is roughly 250 lines containing several nested closures (`create_fresh_connection`, `wrap_execution`, `wrap_cancellation`, `submit_work_item`, `invalidate_connection`, `stream_reader`, `get_reconnect_delay_seconds`). The function captures a lot of mutable state via `nonlocal` and is now central to the worker's correctness on reconnects, drains, and silent disconnects.

The merged dotnet PR (#708) split the equivalent code into `WorkItemStreamConsumer`, `ReconnectBackoff`, and `ProcessorExitReason`. That structure is what makes its behavior-level tests possible without monkey-patching internals.

Suggestion (not necessarily for this PR, but a strong follow-up):

- Extract a `_WorkItemStreamConsumer` class that owns the reader thread, sentinel handling, and outcome classification.
- Move `create_fresh_connection` / `invalidate_connection` to methods on `TaskHubGrpcWorker`.
- Move the in-flight tracker acquire/release wrapping into a small helper class alongside `_InFlightChannelTracker`.

The resulting `_async_run_loop` should read closer to the dotnet processor: a small loop that asks each helper for the next step.

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[7/10] Promote the `30.0`s grace period and `30` jitter cap to named constants**

Two semantically meaningful numbers are hard-coded:

- `durabletask/client.py` — the retired-channel close timer in both sync and async clients:
  ```python
  close_timer = threading.Timer(30.0, self._close_retired_channel, args=(old_channel,))
  ...
  await asyncio.sleep(30.0)
  ```
- `durabletask/internal/grpc_resiliency.py`:
  ```python
  capped_attempt = min(attempt, 30)
  ```

These both encode design choices (in-flight RPC drain window; exponent ceiling to keep `2 ** n` bounded) that mirror the dotnet PR. Promoting them to module-level constants with a short comment makes the intent discoverable to future readers and lets tests reference the same constant.

Suggested names:
- `_RETIRED_CHANNEL_CLOSE_DELAY_SECONDS = 30.0`
- `_MAX_JITTER_ATTEMPT_EXPONENT = 30`

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[9/10] `_InFlightChannelTracker` keys on `id(channel)`; consider `WeakKeyDictionary`**

```python
class _InFlightChannelTracker:
    def __init__(self):
        self._lock = Lock()
        self._states: dict[int, _TrackedChannelState] = {}

    def acquire(self, channel: Any):
        channel_key = id(channel)
        ...
```

This is safe as written because `_TrackedChannelState` holds a strong reference to the channel, so `id()` reuse-after-GC can't bite here. But `id`-keyed dicts are an unusual idiom in Python and are a well-known footgun if anyone later changes the lifetime model (e.g., drops the strong reference in `_TrackedChannelState`). A reader has to verify the invariant by reading the whole class.

Suggestion: either use the channel itself as the key if `grpc.Channel` is hashable (it is, by identity), or use `weakref.WeakKeyDictionary` if you want explicit lifecycle semantics. Both make the invariant local instead of structural.

andystaples:
> _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a code review pass. Treat as a reviewer suggestion, not authoritative._

**[10/10] Worker tests reach deep into private state; consider extracting `_WorkItemStreamConsumer` first**

`tests/durabletask/test_worker_resiliency.py` is ~841 lines and exercises the worker by monkey-patching private attributes:

- `worker._async_worker_manager = DummyWorkerManager()`
- `worker._shutdown.wait = lambda timeout: False`
- `worker._execute_activity = ...`
- `worker._cancel_activity = ...`
- `monkeypatch.setattr("durabletask.worker.shared.get_grpc_channel", ...)`
- `monkeypatch.setattr("durabletask.worker.stubs.TaskHubSidecarServiceStub", ...)`

This mirrors the merged dotnet PR's style and gives strong behavioral coverage today, but it tightly couples the tests to the current shape of `TaskHubGrpcWorker._async_run_loop` and its module-level imports. Any of the follow-ups suggested elsewhere in this review (extracting the stream consumer, switching the client to a `UnaryUnaryClientInterceptor`, renaming `_can_recreate_channel`) will require fan-out edits across these tests.

Suggestion: pair the extraction proposed in comment [6/10] with a rewrite of the largest test cases against the new helper class. The remaining worker-level tests should only need to verify wiring, not transport semantics.

durabletask/worker.py
```diff
@@ -2878,11 +3130,10 @@ def submit_entity_batch(self, func, cancellation_func, *args, **kwargs):
 
     def shutdown(self):
         self._shutdown = True
-        self.thread_pool.shutdown(wait=True)
 
     async def reset_for_new_run(self):
         """Reset the manager state for a new run."""
-        self._shutdown = False
+        self.prepare_for_run()
```
andystaples, line 3137: Nit: prepare_for_run seems to be abstracted from reset_for_new_run only for testing purposes, is this necessary and/or best practice?

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import random
+from dataclasses import dataclass
+
+import grpc
+
+LONG_POLL_METHODS = {"WaitForInstanceStart", "WaitForInstanceCompletion"}
```
andystaples: Nit: Hardcoded list of method names, though I'm not sure if that is preferable vs a string matching approach like WaitFor* - just thinking ahead to when we add new methods that would qualify, do we need to go back and maintain this list

durabletask/client.py
```diff
@@ -166,24 +176,109 @@ def __init__(self, *,
...
         if channel is None:
-            interceptors = prepare_sync_interceptors(metadata, interceptors)
             channel = shared.get_grpc_channel(
-                host_address=host_address,
+                host_address=self._host_address,
                 secure_channel=secure_channel,
-                interceptors=interceptors,
+                interceptors=self._interceptors,
                 channel_options=channel_options,
             )
```
andystaples, line 241: Uncertain: Pass-by-reference - does get_grpc_channel modify the interceptor list or the interceptors, and do we want those changes in the client's copy when the channel is recreated? If yes, disregard

durabletask/internal/grpc_resiliency.py
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Good point — I considered prefix matching but kept the explicit set deliberately. The risk with `WaitFor*` is that a future RPC like `WaitForApproval` or `WaitForExternalEvent` might match the prefix but not actually be a server-side long-poll (i.e., its `DEADLINE_EXCEEDED` really would mean a transport failure). The explicit set forces an intentional decision when a new long-poll method ships, which is cheap because the list is exactly two entries and only changes when the proto changes.

I''ve added a short comment above the constant explaining the semantics. Happy to switch to a prefix-based match if you''d prefer; let me know.

durabletask/worker.py
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Not test-only — `prepare_for_run` is also called from `TaskHubGrpcWorker._async_run_loop` at the start of each run (line 706). The split exists because `reset_for_new_run` additionally drains the queues (cancelling any in-flight items), which is only valid when there''s a running event loop, while `prepare_for_run` is the lightweight "this manager is about to run again" path that just resets `_shutdown` and the thread pool. The resiliency feature relies on the manager being reusable across reconnects, so we need a reset path that doesn''t require draining live queues. The tests call `prepare_for_run` for the same reason — they exercise the manager-reuse path, not a test-only seam.

durabletask/client.py
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Checked — `shared.get_grpc_channel` does not mutate the interceptor sequence. It passes the values through `grpc.intercept_channel(channel, *interceptors)` (unpacking, no in-place modification) for sync, and forwards the sequence directly to `grpc.aio.{secure,insecure}_channel(..., interceptors=...)` on the async side. Neither call retains a write-through reference.

We also defensively copy in the client `__init__` (`self._interceptors = list(resolved_interceptors)`), so even if gRPC ever started mutating the argument, the client''s copy would not be affected — and the same list is reused on every recreate so the recreated channel always gets the same interceptors. So this one''s fine as-is.

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def is_worker_transport_failure(status_code: grpc.StatusCode) -> bool:
+    return status_code in {
+        grpc.StatusCode.UNAVAILABLE,
+        grpc.StatusCode.DEADLINE_EXCEEDED,
+    }
+
+
+class ClientResiliencyInterceptor(grpc.UnaryUnaryClientInterceptor):
```
berndverst, line 87: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[1/10] Centralize resiliency in a `UnaryUnaryClientInterceptor` instead of `_invoke_unary`](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482609760)

**Done.** Implemented `ClientResiliencyInterceptor` (and `AsyncClientResiliencyInterceptor`) here, wired them into `TaskHubGrpcClient` / `AsyncTaskHubGrpcClient`, removed both `_invoke_unary` methods, and reverted all 34 stub call sites to ordinary `self._stub.MethodName(req)`. `LONG_POLL_METHODS` and `is_client_transport_failure` now operate on full gRPC method paths so the interceptor can match `ClientCallDetails.method` directly. Test shims (`_ResilientSyncTestStub`/`_ResilientAsyncTestStub` + `install_resilient_test_stubs`) keep the existing tests that patch `stubs.TaskHubSidecarServiceStub` participating in the interceptor pipeline. Caller-owned channels deliberately bypass the interceptor (they're never recreated, and `grpc.aio` has no public `intercept_channel` equivalent).

durabletask/worker.py
```diff
@@ -448,9 +536,18 @@ def __init__(
         self._shutdown = Event()
         self._is_running = False
         self._channel = channel
+        # The SDK owns (and may recreate) the gRPC channel only when the caller
+        # did not provide one. Mirrors ``TaskHubGrpcClient._owns_channel`` so
+        # both files use the same name for the same concept.
+        self._owns_channel = channel is None
```
berndverst, line 542: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[2/10] Align naming for "SDK-owned channel"](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482610070)

**Done.** Added `self._owns_channel = channel is None` to `TaskHubGrpcWorker.__init__` (mirroring the clients) and removed `_can_recreate_channel()`. All four call sites and the worker test now use `self._owns_channel` directly.

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,153 @@
...
+        base_seconds: float,
+        cap_seconds: float,
+) -> float:
+    capped_attempt = min(attempt, _MAX_JITTER_ATTEMPT_EXPONENT)
+    upper_bound = min(cap_seconds, base_seconds * (2 ** capped_attempt))
+    return random.random() * upper_bound
+
+
+@dataclass
+class FailureTracker:
```
berndverst, line 39: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[3/10] `FailureTracker` thread safety](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482610462)

**Done.** `FailureTracker` now owns an internal `threading.Lock` and `record_failure` / `record_success` execute their read-modify-write under it. Async callers pay a negligible cost because they only touch the tracker from the event loop, but the sync path is no longer racy. Added a regression test that fires 8 threads x 500 increments and asserts the counter ends at 4000.

durabletask/worker.py
```diff
@@ -2648,11 +2887,27 @@ def __init__(self, concurrency_options: ConcurrencyOptions, logger: logging.Logg
         self._pending_activity_work: list = []
         self._pending_orchestration_work: list = []
         self._pending_entity_batch_work: list = []
-        self.thread_pool = ThreadPoolExecutor(
-            max_workers=concurrency_options.maximum_thread_pool_workers,
+        self.thread_pool = self._create_thread_pool()
+        self._pool_is_shutdown = False
```
berndverst, line 2891: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[4/10] Don't depend on `ThreadPoolExecutor._shutdown`](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482610774)

**Done.** `_AsyncWorkerManager` now tracks an explicit `self._pool_is_shutdown` flag, set in `__init__`, reset in `_ensure_thread_pool`, and flipped to `True` in `run()`'s `finally` after `shutdown(wait=True)`. The two `getattr(self.thread_pool, "_shutdown", False)` reads are gone. Added a test that runs the manager once, then `prepare_for_run`s and asserts a fresh pool was created.

durabletask/worker.py
```diff
@@ -590,44 +707,81 @@ def run_loop():
...
                 conn_retry_count = 0
                 self._logger.info(f"Created fresh connection to {self._host_address}")
             except Exception as e:
                 self._logger.warning(f"Failed to create connection: {e}")
-                current_channel = self._channel if self._channel is not None else None
                 current_stub = None
                 raise
 
-        def invalidate_connection():
+        def wrap_with_release(handler, release):
```
berndverst, line 749: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[5/10] Collapse `wrap_execution` / `wrap_cancellation`](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482611078)

**Done.** Replaced with a single `wrap_with_release(handler, release)` helper used for both the execution and cancellation paths in `submit_work_item`.

durabletask/worker.py
```diff
@@ -590,44 +707,81 @@ def run_loop():
         self._is_running = True
 
     async def _async_run_loop(self):
```
berndverst, line 709: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[6/10] Extract `_WorkItemStreamConsumer`](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482611383)

**Deferring** per your "not necessarily for this PR" note. Agree the long-term shape should mirror the .NET `WorkItemStreamConsumer` / `ReconnectBackoff` / `ProcessorExitReason` split, and that will make the worker tests in [10/10] much smaller. I'll track it together with the rest of the worker refactor as a follow-up.

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,153 @@
...
+# expiring rather than a transport failure, so we do not treat it as one.
+LONG_POLL_METHODS = {
+    "/TaskHubSidecarService/WaitForInstanceStart",
+    "/TaskHubSidecarService/WaitForInstanceCompletion",
+}
+
+# Cap the attempt number fed into ``2 ** attempt`` to keep the jitter calculation
+# bounded for callers that retry indefinitely; once we hit the cap, the upper
+# bound is fully governed by ``cap_seconds``.
+_MAX_JITTER_ATTEMPT_EXPONENT = 30
```
berndverst, line 24: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[7/10] Promote `30.0`s grace period and `30` jitter cap to named constants](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482611699)

**Done.** Added `_RETIRED_CHANNEL_CLOSE_DELAY_SECONDS = 30.0` at module scope in `durabletask/client.py` (used by both the sync `threading.Timer` and the async `asyncio.sleep`) and `_MAX_JITTER_ATTEMPT_EXPONENT = 30` in `durabletask/internal/grpc_resiliency.py`, each with a short comment explaining the intent.

durabletask/worker.py
```diff
@@ -115,6 +123,83 @@ class VersionFailureStrategy(Enum):
...
+
+
+@dataclass
+class _TrackedChannelState:
+    channel: Any
+    ref_count: int = 0
+    close_when_released: bool = False
+
+
+class _InFlightChannelTracker:
```
berndverst, line 140: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[9/10] `_InFlightChannelTracker` keying](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482612490)

**Done.** The tracker now keys directly on the channel object (`dict[Any, _TrackedChannelState]`) instead of `id(channel)`. gRPC channels are hashable by identity, and `_TrackedChannelState` continues to hold the strong reference, so the lifetime invariant is now local to the class and a reader doesn't have to chase the strong-reference argument.

tests/durabletask/test_worker_resiliency.py
```diff
@@ -0,0 +1,861 @@
...
+from unittest.mock import MagicMock
+
+import pytest
+
+from durabletask.grpc_options import GrpcWorkerResiliencyOptions
+from durabletask.internal import orchestrator_service_pb2 as pb
+from durabletask.worker import (
+    _AsyncWorkerManager,
+    ConcurrencyOptions,
+    TaskHubGrpcWorker,
```
berndverst, line 13: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re: [[10/10] Worker tests reach deep into private state](https://github.com/microsoft/durabletask-python/pull/135#issuecomment-4482612802)

**Deferring** with [6/10]. Once the stream consumer is extracted, the largest blocks in `test_worker_resiliency.py` should collapse into thin wiring checks plus consumer-level behavior tests, and we won't need the `_async_worker_manager` / `_execute_activity` / `get_grpc_channel` monkey-patching.

durabletask-azuremanaged/durabletask/azuremanaged/client.py
```diff
@@ -104,6 +111,7 @@ def __init__(self, *,
                  secure_channel: bool = True,
                  interceptors: Optional[Sequence[shared.AsyncClientInterceptor]] = None,
                  channel_options: Optional[GrpcChannelOptions] = None,
+                 resiliency_options: Optional[GrpcClientResiliencyOptions] = None,
```
berndverst, line 114: Does this not break existing code that uses positional arguments? Maybe we should move these new options to be the last parameter.

durabletask-azuremanaged/durabletask/azuremanaged/worker.py
```diff
@@ -74,6 +79,7 @@ def __init__(self, *,
                  secure_channel: bool = True,
                  interceptors: Optional[Sequence[shared.ClientInterceptor]] = None,
                  channel_options: Optional[GrpcChannelOptions] = None,
+                 resiliency_options: Optional[GrpcWorkerResiliencyOptions] = None,
```
berndverst, line 82: Consider positional argument placement.

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,153 @@
...
+    async def _record_outcome(self, method: str, error: Optional[BaseException]) -> None:
+        if error is None:
+            self._failure_tracker.record_success()
+            return
+        status_code = getattr(error, "code", lambda: None)()
+        if status_code is not None and is_client_transport_failure(method, status_code):
+            if self._failure_tracker.record_failure():
+                result = self._on_recreate()
+                if hasattr(result, "__await__"):
+                    await result
```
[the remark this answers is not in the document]
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Done in commit 935503f. Added `import inspect`, switched the check to `inspect.isawaitable(result)`, and tightened the `on_recreate` callback type annotation to `Callable[[], Union[None, Awaitable[object]]]` so the contract (sync callbacks return `None`; async callbacks return an `Awaitable[object]` that we await) is explicit.

durabletask-azuremanaged/durabletask/azuremanaged/client.py
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

No breakage: `DurableTaskSchedulerClient.__init__` and `AsyncDurableTaskSchedulerClient.__init__` both use `def __init__(self, *, ...)` — the leading `*` makes every parameter keyword-only, so positional-arg callers cannot exist. The same is true on the base `TaskHubGrpcClient` / `AsyncTaskHubGrpcClient` (also keyword-only), and the `super().__init__(...)` calls inside the azuremanaged subclasses pass everything by keyword. The placement is therefore purely a style choice; I left `resiliency_options` next to the related `channel_options` for readability, but happy to move it to the end if you prefer.

durabletask-azuremanaged/durabletask/azuremanaged/worker.py
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Same answer as on the client side: `DurableTaskSchedulerWorker.__init__` is `def __init__(self, *, ...)` (keyword-only), and so is the base `TaskHubGrpcWorker.__init__`, so adding `resiliency_options` in the middle cannot break any positional caller. The `super().__init__(...)` call below also passes everything by keyword. Placement is purely stylistic; I grouped it with the related `channel_options` for readability, but happy to move it to the end if you prefer.

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,154 @@
...
+    async def _record_outcome(self, method: str, error: Optional[BaseException]) -> None:
+        if error is None:
+            self._failure_tracker.record_success()
+            return
+        status_code = getattr(error, "code", lambda: None)()
+        if status_code is not None and is_client_transport_failure(method, status_code):
+            if self._failure_tracker.record_failure():
+                result = self._on_recreate()
+                if inspect.isawaitable(result):
+                    await result
```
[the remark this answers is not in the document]
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Re-evaluated: the warning is a false positive — `await result` is never side-effect-free here; the whole purpose of the line is to *run* the async recreate callback to completion (suspend, schedule, await, propagate exceptions). CodeQL's `py/ineffectual-statement` heuristic doesn't model that.

Rather than dismiss, I made the intent explicit in commit 053b579: rewrote the line as `_ = await result` with a short comment. That keeps the same runtime behaviour and is the idiomatic Python signal for "awaiting purely for the side effect; return value intentionally discarded".

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,157 @@
...
+            return
+        status_code = getattr(error, "code", lambda: None)()
+        if status_code is not None and is_client_transport_failure(method, status_code):
+            if self._failure_tracker.record_failure():
+                result = self._on_recreate()
+                if inspect.isawaitable(result):
+                    # ``_ =`` signals that we await purely for the side effect
+                    # of running the recreate callback to completion; the
+                    # awaitable's return value is intentionally discarded.
+                    _ = await result
```
andystaples: > _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a follow-up review pass on the interceptor refactor. Treat as a reviewer suggestion, not authoritative._

**Follow-up [1/3] — Channel recreate runs on the caller's RPC-failure path**

Nice refactor on [1/10] — the interceptor is much cleaner than `_invoke_unary`. One regression worth flagging though: the recreate work is no longer kept off the caller's failure path the way the dotnet PR does it.

**dotnet** (`ChannelRecreatingCallInvoker.MaybeTriggerRecreate`):

```csharp
// Keep recreate work off the caller's RPC-failure path.
_ = Task.Run(() => this.RecreateAsync(observedCount));
```

**this PR** (`AsyncClientResiliencyInterceptor.intercept_unary_unary`):

```python
try:
    response = await continuation(client_call_details, request)
except grpc.aio.AioRpcError as rpc_error:
    await self._record_outcome(client_call_details.method, rpc_error)   # <-- awaits recreate
    raise
```

…and `_record_outcome` ultimately does `_ = await result` on `self._on_recreate()`, which is `_maybe_recreate_channel()`. That acquires `self._recreate_lock`, calls `shared.get_async_grpc_channel(...)` (DNS + TLS handshake on the new channel), constructs a new stub, then returns — and only *then* does the original `AioRpcError` propagate to the awaiting caller.

The sync side has the same issue and is arguably worse: the interceptor runs on the calling thread, so `self._stub.GetInstance(req)` blocks for `failed RPC + lock wait + channel creation` before the exception surfaces. During a real outage (slow DNS, TLS handshake to a recovering peer), that's seconds of caller-visible extra latency, and it serializes across threads on `_recreate_lock`.

Suggested fix:

- **Async:** in `AsyncClientResiliencyInterceptor._record_outcome`, replace the `await result` with `asyncio.create_task(result)` (still need to keep a reference so the task isn't GC'd; the same `_retired_channel_close_tasks`-style set works).
- **Sync:** in `ClientResiliencyInterceptor._record_outcome`, fire `_maybe_recreate_channel` on a dedicated single-flight `threading.Thread` (daemon=True) or a 1-worker `ThreadPoolExecutor` reused across calls. The existing `_recreate_lock` + `min_recreate_interval_seconds` cooldown already prevents thrash, so a single-flight worker thread is enough.

Behavior should match the dotnet design: triggering call still gets its original error promptly; *subsequent* calls benefit from the recreated channel.


---

_Part of a 3-comment follow-up review:_
- [1/3] Recreate on RPC failure path
- [2/3] `__init__` constructs interceptor before its state
- [3/3] Async interceptor: narrow `except` + original-error swallow

_Tightly related to follow-up [3/3]: moving recreate to fire-and-forget also resolves the original-error swallow risk._

andystaples (commented):
Three items - large refactors look good, just a few remaining items to think about

durabletask/client.py
```diff
@@ -166,24 +183,110 @@ def __init__(self, *,
...
+        # resiliency interceptor: the caller controls the channel lifetime and
+        # we never recreate it, so failure tracking against it would have no
+        # observable effect. Callers wanting resiliency on a custom channel
+        # can prepend the interceptor themselves via grpc.intercept_channel.
         self._channel = channel
         self._stub = stubs.TaskHubSidecarServiceStub(channel)
+        self._closing = False
+        self._last_recreate_time = 0.0
+        self._recreate_lock = threading.Lock()
+        self._retired_channels: dict[grpc.Channel, threading.Timer] = {}
```
andystaples: > _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a follow-up review pass on the interceptor refactor. Treat as a reviewer suggestion, not authoritative._

**Follow-up [2/3] — `__init__` constructs the interceptor before the state it depends on**

In both `TaskHubGrpcClient.__init__` and `AsyncTaskHubGrpcClient.__init__`, the resiliency interceptor is constructed with a bound reference to `self._maybe_recreate_channel` *before* the attributes that method reads (`self._closing`, `self._recreate_lock`, `self._last_recreate_time`, `self._retired_channels` / `_retired_channel_close_tasks`) are assigned.

Sync example:

```python
self._client_failure_tracker = FailureTracker(...)
self._resiliency_interceptor = ClientResiliencyInterceptor(
    self._client_failure_tracker,
    self._maybe_recreate_channel,        # captures bound method now
)
...
if channel is None:
    channel = shared.get_grpc_channel(..., interceptors=self._interceptors, ...)
self._channel = channel
self._stub = stubs.TaskHubSidecarServiceStub(channel)
self._closing = False                     # <-- not set yet when bound method was captured
self._last_recreate_time = 0.0
self._recreate_lock = threading.Lock()
self._retired_channels: dict[...] = {}
```

Today this works because `shared.get_grpc_channel(...)` doesn't fire any RPCs — but it's a fragile invariant: anyone adding an eager handshake-style call in `get_grpc_channel`, or any test that exercises the interceptor immediately after construction, would `AttributeError` on `self._closing`.

Suggested fix: hoist the state attributes above the interceptor construction (and above the channel creation, since both interceptors need the lock and `_closing` to be defined to be safe to call):

```python
self._closing = False
self._last_recreate_time = 0.0
self._recreate_lock = threading.Lock()
self._retired_channels: dict[grpc.Channel, threading.Timer] = {}
self._client_failure_tracker = FailureTracker(...)
self._resiliency_interceptor = ClientResiliencyInterceptor(
    self._client_failure_tracker, self._maybe_recreate_channel,
)
```

Same change applies symmetrically in `AsyncTaskHubGrpcClient.__init__`.


---

_Part of a 3-comment follow-up review:_
- [1/3] Recreate on RPC failure path
- [2/3] `__init__` constructs interceptor before its state
- [3/3] Async interceptor: narrow `except` + original-error swallow

durabletask/internal/grpc_resiliency.py
```diff
@@ -0,0 +1,157 @@
...
+        self._on_recreate = on_recreate
+
+    async def intercept_unary_unary(self, continuation, client_call_details, request):
+        try:
+            response = await continuation(client_call_details, request)
+        except grpc.aio.AioRpcError as rpc_error:
+            await self._record_outcome(client_call_details.method, rpc_error)
+            raise
+        await self._record_outcome(client_call_details.method, None)
+        return response
```
andystaples, line 163: > _🤖 The following comment was generated by an AI agent (GitHub Copilot CLI) as part of a follow-up review pass on the interceptor refactor. Treat as a reviewer suggestion, not authoritative._

**Follow-up [3/3] — Async interceptor: narrow `except` + potential original-error swallow**

Two related small issues in `AsyncClientResiliencyInterceptor.intercept_unary_unary`:

```python
async def intercept_unary_unary(self, continuation, client_call_details, request):
    try:
        response = await continuation(client_call_details, request)
    except grpc.aio.AioRpcError as rpc_error:
        await self._record_outcome(client_call_details.method, rpc_error)
        raise
    await self._record_outcome(client_call_details.method, None)
    return response
```

1. **Narrow `except`.** Only `grpc.aio.AioRpcError` is caught. Cancellation (`asyncio.CancelledError`), timeouts, and programming errors bypass `_record_outcome` entirely, so the failure tracker does not see them and does not reset on success-after-non-RpcError either. The sync interceptor side-steps this by reading `.exception()` off the call object (which includes non-RpcError outcomes); the async path doesn't have that affordance. Consider:

   ```python
   except BaseException as exc:
       if isinstance(exc, grpc.aio.AioRpcError):
           await self._record_outcome(client_call_details.method, exc)
       # Don't count cancellation / programming errors, but do reset
       # the failure counter on any non-transport outcome — same policy
       # as the sync side.
       else:
           self._failure_tracker.record_success()
       raise
   ```

2. **Original error can be replaced if `_record_outcome` raises.** If the user-supplied `on_recreate` callback raises (or `asyncio.create_task` raises with no running loop, etc.), the new exception propagates instead of the original `AioRpcError`. Wrapping the body of `_record_outcome` in a `try/except` that logs and swallows, or moving recreate to a fire-and-forget task per [follow-up 1/3], both fix this — the second one fixes it for free.


---

_Part of a 3-comment follow-up review:_
- [1/3] Recreate on RPC failure path
- [2/3] `__init__` constructs interceptor before its state
- [3/3] Async interceptor: narrow `except` + original-error swallow

_See follow-up [1/3] for the fire-and-forget fix that also closes the original-error swallow._

durabletask/internal/grpc_resiliency.py
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Fair point — agreed, and addressed in [`3268b45`](https://github.com/microsoft/durabletask-python/commit/3268b452a8d65f10c4fc1b0e8e4b8eaffa9e7842).

Both interceptors now hand off the recreate to the client via a synchronous `_schedule_recreate` callback that fires-and-forgets the actual work:

- **Sync**: `_schedule_recreate` spawns a daemon `threading.Thread` (single-flight under `_recreate_thread_lock`); the failing RPC's original error propagates immediately. `close()` joins the recreate thread with a 5 s bound so teardown stays deterministic.
- **Async**: `_schedule_recreate` does `asyncio.create_task(self._run_recreate())`, with the task stored on the client (so it isn't GC'd) and single-flighted via `_recreate_task.done()`. `close()` awaits any in-flight recreate task before tearing the channel down.

The existing `min_recreate_interval_seconds` cooldown still gates the actual channel construction inside `_maybe_recreate_channel`, so the single-flight guard is purely about not spawning redundant daemon threads / tasks while one is in flight.

A `_recreate_done_event` (`threading.Event` / `asyncio.Event`) is set in `_run_recreate`'s `finally` so tests can synchronise on completion without polling — used in the handful of tests in `test_client.py` that previously relied on the recreate running inline.

You're right that this isn't strictly a regression from the interceptor refactor (`_invoke_unary` did the same thing pre-refactor), but the centralised location in the interceptor made it the natural place to fix and brings us closer in shape to the .NET client. Good catch.

durabletask/client.py
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Good catch — fixed in [`3268b45`](https://github.com/microsoft/durabletask-python/commit/3268b452a8d65f10c4fc1b0e8e4b8eaffa9e7842).

The resiliency state (`_closing`, `_recreate_lock`, `_last_recreate_time`, `_retired_channels` / `_retired_channel_close_tasks`) is now initialised before `ClientResiliencyInterceptor` / `AsyncClientResiliencyInterceptor` is constructed in both `TaskHubGrpcClient.__init__` and `AsyncTaskHubGrpcClient.__init__`. Pure reordering, no behavior change in practice today, but it removes the latent footgun where any failure handled during construction of the underlying channel could observe a half-built client. Also added the new `_recreate_thread_lock` / `_recreate_thread` (sync), `_recreate_task` (async), and `_recreate_done_event` next to the existing state so everything the interceptor's bound callback depends on is set up together.

durabletask/internal/grpc_resiliency.py
berndverst: > _Reply generated by GitHub Copilot CLI on behalf of @berndverst._

Both sub-issues addressed in [`3268b45`](https://github.com/microsoft/durabletask-python/commit/3268b452a8d65f10c4fc1b0e8e4b8eaffa9e7842):

**Sub-issue A (narrow except clause):** The async interceptor now catches `Exception` rather than `grpc.aio.AioRpcError`:

```python
except Exception as exc:
    if isinstance(exc, grpc.aio.AioRpcError):
        self._record_outcome(client_call_details.method, exc)
    else:
        self._failure_tracker.record_success()
    raise
```

Using `Exception` (not `BaseException`) is intentional: `asyncio.CancelledError`, `KeyboardInterrupt`, and `SystemExit` propagate without bookkeeping, which is the correct asyncio convention. Non-`AioRpcError` `Exception`s reset the failure counter to mirror the sync interceptor's policy (in sync, `response.exception()` returning a non-RpcError falls through to `record_success`).

**Sub-issue B (original-error replacement):** Resolved automatically by the fire-and-forget refactor for [1/3]. `_record_outcome` is now a synchronous method on both interceptors — no `await` on the exception path, and `_schedule_recreate` wraps its body in a `try / except Exception` that logs and swallows any scheduling failure. The caller's `AioRpcError` is guaranteed to propagate to `raise` unchanged.

andystaples (approved):

Issue #137, opened 2026-05-19
Potential malicious code execution in durabletask versions 1.4.1, 1.4.2, and 1.4.3


BarElrom-UpWind:
Hi,
I’m reporting a suspected supply chain compromise affecting `durabletask 1.4.1, 1.4.2, 1.4.3`

These versions include import-time code that downloads and executes a remote payload: `https://check.git-service[.]com/rope.pyz`
Version details:
```
Version	Upload UTC	Injected files
1.4.1	        16:19            	__init__.py
1.4.2	16:49        	__init__.py, task.py
1.4.3	16:54          	__init__.py, task.py, entities/__init__.py, extensions/__init__.py, payload/__init__.py
```

IOCs
Payload SHA256: 069ac1dc7f7649b76bc72a11ac700f373804bfd81dab7e561157b703999f44ce
Payload URL: https://check.git-service[.]com/rope.pyz
C2/fallback: t.m-kosche[.]com

Please urgently investigate, yank the affected PyPI versions, publish a clean version, and advise users to rotate secrets and inspect systems where these versions were installed or imported.

KunalSin9h:
We are also tracking it, Confirmed Supply chain attack.  

Technical Details: https://safedep.io/malicious-durabletask-pypi-supply-chain-attack/

torosent:
The packages are yanked and we are working on the rest of the mitigation

BarElrom-UpWind:
Confirmed - It's supply chain attack:
https://www.upwind.io/feed/newly-discovered-durabletask-malware-targeted-kubernetes-cloud-secrets-and-ci-cd-infrastructure

ashishkurmi:
Thanks to @torosent and the maintainers for the swift response in yanking the affected packages.
We have shared our findings here: https://www.stepsecurity.io/blog/microsofts-durabletask-pypi-package-compromised-in-supply-chain-attack

berndverst:
The `durabletask` project was automatically delisted / quarantined from PyPI, and out of precaution the associated maintainer accounts were suspended by PyPI.

We are working with the PyPI security team to relist (remove from quarantine) the project and restore our PyPI access.

Thanks to the PyPI security team for the help!

berndverst:
The project access has now been relisted and removed from quarantine. The latest release [1.4.0](https://pypi.org/project/durabletask/1.4.0/) is safe for use.

Thanks again to the PyPI security team for the prompt assistance.

[...]

PR #140, opened 2026-05-20
Use Trusted Publishing in CI package upload to PyPI

### LINKED ISSUE #139: Switch to Trusted Publishing for package upload to PyPI in CI
> Following the breach (#137) in which a PyPI credential was stolen and used to republish a malicious package, the appropriate remedy is to remove the use of PyPI credentials and switch to Trusted Publishing.

Trusted publishing (with attestations) is currently [suggested](https://docs.pypi.org/trusted-publishers/) as a security practice to ensure confidence that what users download from PyPI is the same artifact that was generated in GitHub CI, meaning that what I see in GitHub is the same as what is installed - handy for auditing. It also avoids the presence of long-lived PyPI tokens as 'secrets' in the CI, which can be exfiltrated by bad actors (as has occurred).

https://github.com/microsoft/durabletask-python/blob/e2f0470614ac5a51324309fe241b906f1074332c/.github/workflows/durabletask.yml#L118-L123

Following the recent supply chain attack on this package, I would strongly advise this code be changed and the stored PyPI API token/GitHub secret removed, not merely rotated.

Instead of using the stored secret token, GitHub vouches for the upload, and PyPI checks the uploading CI workflow filename against a pre-declared filename you set ahead of time (in durabletask’s case the filename is `durabletask.yml`). It also requires creating an 'environment' for the repo on GitHub (suggested name "pypi") which you register the package with on PyPI.

### Steps to implement:

> Instead of installing twine directly in the CI runner, use the [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) action which handles the Trusted Publishing for you

* Register the PyPI package for Trusted Publishing

* Add the [environment definition](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idenvironment) to the `publish-release` job

* Add `id-token: write`, `contents: write` and `packages: write` [permissions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idpermissions) to the `publish-release` job
  
  * specifying any permission restricts the token, and creating a GitHub release requires `contents: write` and potentially `packages: write`

* Remove `user` and `password` arguments in the `publish-release` job

* Remove the `PYPI_API_TOKEN` project secret from both the code and the secret store

lmmx:
- Proposed changes to resolve #139
  - Follows #137

### DIFF
--- a/.github/workflows/durabletask.yml
+++ b/.github/workflows/durabletask.yml
@@ -90,34 +90,29 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/v')  # Only run if a matching tag is pushed
     needs: run-tests
     runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      id-token: write
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
 
       - name: Extract version from tag
-        run: echo "VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_ENV  # Extract version from the tag
+        run: echo "VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_ENV
 
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.14"  # Adjust Python version as needed
+          python-version: "3.14"
 
-      - name: Install dependencies
+      - name: Install build dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install build twine
+          pip install build
 
       - name: Build package from root directory
         run: |
           python -m build
 
-      - name: Check package
-        run: |
-          twine check dist/* 
-
       - name: Publish package to PyPI
-        env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN  }}  # Store your PyPI API token in GitHub Secrets
-        run: | 
-          twine upload dist/*
+        uses: pypa/gh-action-pypi-publish@release/v1


lmmx:
@microsoft-github-policy-service agree

.github/workflows/durabletask.yml
```diff
@@ -90,34 +90,29 @@ jobs:
...
-      - name: Install dependencies
+      - name: Install build dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install build twine
+          pip install build
 
       - name: Build package from root directory
         run: |
           python -m build
```
webknjaz, line 115: Do not run this with OIDC privileges. Move to a separate job w/ limited permissions.

https://github.com/zizmorcore/zizmor/issues/1968

berndverst:
We are migrating to trusted publishing to occur outside of GitHub Actions using a unified Microsoft publisher account. Thanks for the PR, however we cannot accept any iteration of this work - in fact we will remove our publishing pipeline from GitHub pipelines entirely.

webknjaz:
@berndverst that's not what trusted publishing means 🤦‍♂️

Sounds like you're still going to be using long-living tokens, which is besides being a bad idea also won't allow you to have digital signatures published to PyPI.

berndverst:
@webknjaz with all due respect - I do not appreciate that response. I know what trusted publishing is and we are adopting it. We are not using any PyPI publishing tokens anymore. All of this work is happening elsewhere within Microsoft managed infrastructure outside of GitHub via a secure centralized process going forward however. This centralized system naturally uses OpenID Connect authentication.

webknjaz:
@berndverst apologies for the tone.

Trusted Publishing isn't just OIDC, though. It's a mechanism built on top, yes.
PyPI, however, won't mark uploads as coming from a trusted platform unless said platform is explicitly set up as trusted. Public GitHub Actions (github.com) is such a platform but anything self-hosted / behind firewall / etc. isn't.

Do you mean that you'll be using one of Google Cloud, Active State or GitLab, then? Here's the list of platforms that will give you trusted identities: https://docs.pypi.org/trusted-publishers/adding-a-publisher/.

Here's some explanation of why arbitrary identity providers are untrusted: https://discuss.python.org/t/new-oidc-providers-for-trusted-publishing/106334.

Here, Seth formalized the ideas that other indices picked up: https://repos.openssf.org/trusted-publishers-for-all-package-repositories.html – a trusted identity can be used to get a short-lived token but tokens are still involved.

So I guess I'm trying to understand what you mean. Are you going to stop publishing to PyPI? Are you going to somehow fake the identity of the public GH repo within your infra behind the firewall so PyPI would be tricked into thinking it's coming from GHA?

berndverst:
@webknjaz - I appreciate the apology :)

There is a standard Microsoft process that we must follow. Our project was non-compliant and that put us at risk as we saw yesterday - it was in our backlog to onboard to the centralized process.

Take a look at the packages published by user `microsoft`: https://pypi.org/user/microsoft/

Going forward, this will be the exclusive owner account of the `durabletask` package. That account itself uses OIDC + a bunch of other Microsoft internal layers for managing publishing permissions and approvals.

There is no room for us to customize any aspect of this unfortunately. These systems and process are established by several centralized teams for the entire company. I'm sure you understand. Some of the owners of this process are the Python team at Microsoft which includes several Python core devs / maintainers.

webknjaz:
Oh, I see.. This means that you won't be able to ship PEP 740 digital attestations, then since OIDC isn't used for auth against PyPI and is internal. Here's a static site that tracks the adoption: https://trailofbits.github.io/are-we-pep740-yet/.

JFYI, here's where you can see if a specific upload was done via trusted publishing: https://pypi.org/project/msgraph-sdk/#msgraph_sdk-1.58.0-py3-none-any.whl shows that it wasn't.
So you probably have a service that holds a long-living tokens (macaroons) and scopes per upload. This is okay, except for the lack of transparency (and no access to TP and digital attestations).

When TP is set up, it's scoped very tightly to the project and it's source repo — it's not bound to user accounts in PyPI, unlike tokens. TP can also be used in orgs.
By the way, I'm surprised you didn't put the projects under an org. Check this out: https://docs.pypi.org/organization-accounts/ — orgs give some more flexibility.

---

As for using true TP in your case, I imagine it could be possible with some trickery (forwarding minted tokens from GHA to your service on request).

berndverst:
Builds and releases for us will occur in our internal Azure DevOps pipelines that mirror the code.

No release will be triggered from GitHub.

I will reach out to the internal centralized release team to see if they can investigate adopting trusted publishing.

I don't know the history for why the project didn't start out under an organizational account. And quite frankly, I have a lot of higher importance work to get done than pondering this question 🙃 

Working on this repo / package is less than 5% of my job after all.

webknjaz:
Okay, feel free to reach out if there's any TP / org questions that need answering.

berndverst:
I spoke with some Python core devs / maintainers internally who are part of the Microsoft wide Python team. Evidently trusted publishing is not considered inherently more secure and PyPI makes no such claims. The Microsoft publisher account should be considered trusted and Microsoft uses secure internal processes. The internal team however has investigated trusted publishing and will continue to evaluate when / if it makes sense.

I will not be able to relay any further information about this. And this is also out of scope for myself and the maintainers of the DurableTask repositories.

webknjaz:
Yep, your setup sounds secure enough. I won't claim that TP is more secure due to your internal security controls. TP, however, brings transparency to what's uploaded to PyPI for external consumers. When someone sees that a release was made via TP, that implies a certain amount of trust as the build process can be audited and people would know that the publisher didn't inject anything secretly. The digital attestations are tracked through Sigstore transparency entries.

So your process is secure from your PoV but to external observer, this doesn't prove what you say. Subsequently, external users cannot rely on the verification process available through attestation validation since there's nothing to validate.
Having digital attestations is one of those breadcrumbs that also contribute to any future breach investigations.

PR #151, opened 2026-06-09
Add Azure Managed on-demand sandbox preview APIs

YunchuWang:
## The scenario

A Python app using Durable Task Scheduler should be able to keep its normal orchestration worker local, but run selected activities in a DTS-started sandbox worker container.

The intended flow is the same shape as the .NET preview:

1. The local app registers orchestrators and local activities with the existing Azure Managed worker/client APIs.
2. The local app declares a sandbox worker profile: activity names, container image, provider kind, CPU/memory, managed identities, environment variables, and concurrency.
3. DTS stores that declaration and later starts a sandbox worker container from the declared image.
4. The sandbox worker process imports the preview sandbox package, connects back to DTS, registers the activities it can run, and reports activity capacity through the sandbox service.
5. DTS routes only the declared sandbox activity work to that live sandbox worker.

In the sample, the main Python app owns the orchestration and local work; the separate sandbox worker image owns the remote activity implementation.

## What's missing

The Python preview surface needed to line up with the renamed protobuf and .NET SDK surface.

The earlier naming still exposed `on_demand_sandbox` concepts and `substrate` terminology. That made the Python API read differently from the shared contract and from the .NET packages, even though all three are describing the same DTS sandbox worker model.

Reviewers also needed one clear place to understand the Python shape: which package an app imports, which API declares profiles, which worker type runs in the sandbox image, and how the generated protobuf names map to the public API.

## The change

The preview package is now `durabletask.azuremanaged.preview.sandboxes`.

A good review path is:

1. Start with `durabletask-azuremanaged/durabletask/azuremanaged/internal/sandbox_service_pb2*`. These files are regenerated from `sandbox_service.proto` and expose `SandboxActivities`, `SandboxProviderKind`, and `sandbox_provider`.
2. Read `durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/`. This is the public preview API: `SandboxActivitiesClient`, `SandboxWorker`, `SandboxWorkerProfileOptions`, and `sandbox_worker_profile`.
3. Read `examples/sandboxes/` last. It shows the intended app split between the main orchestration process and the sandbox worker process, using the renamed `DTS_SANDBOX_*` settings.

The old `preview.on_demand_sandbox` package and old example path are replaced by the `sandboxes` naming so the Python PR matches the protobuf and .NET PRs.


durabletask-azuremanaged/durabletask/azuremanaged/preview/on_demand_sandbox/client.py
```diff
@@ -0,0 +1,382 @@
...
+        if callable(configure):
+            configure(options)
+
+        _worker_profiles[normalized_profile] = options
+        return cls
+
+    return decorator
+
+
+def resolve_activity_names(activity_names: str | Iterable[str]) -> list[str]:
```
YunchuWang: check if this should be internal function

durabletask-azuremanaged/durabletask/azuremanaged/preview/on_demand_sandbox/client.py
```diff
@@ -0,0 +1,382 @@
...
+    return declaration
+
+
+def build_profile_on_demand_sandbox_activity_declarations() -> list[pb.OnDemandSandboxActivityDeclaration]:
+    """Build on-demand sandbox declarations from worker profile configuration."""
+    declarations: list[pb.OnDemandSandboxActivityDeclaration] = []
+    activity_owners: dict[str, str] = {}
+    for profile in _worker_profiles.values():
+        activity_names = resolve_activity_names(profile.activity_names)
+        if not activity_names:
```
YunchuWang: This should not silently skip a profile with no activity names. Let's fail during profile declaration after `configure(options)` (or at least remove this guard so `build_on_demand_sandbox_activity_declaration(...)` raises) instead of allowing a misconfigured profile to be ignored.

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/client.py
```diff
@@ -0,0 +1,382 @@
+# Copyright (c) Microsoft Corporation.
```
YunchuWang, line 1: Let's keep the preview public surface limited to the exported customer entry points. The module-level helpers like `resolve_activity_names`, `build_on_demand_sandbox_activity_declaration`, `build_profile_on_demand_sandbox_activity_declarations`, `build_on_demand_sandbox_worker_start`, and `build_on_demand_sandbox_worker_heartbeat` should be private, and `connect_on_demand_sandbox_activity_worker(...)` should probably be private too unless customers are expected to use raw protobuf worker streams directly.

durabletask-azuremanaged/durabletask/azuremanaged/preview/on_demand_sandbox/client.py
```diff
@@ -0,0 +1,382 @@
...
+    if not taskhub or not taskhub.strip():
+        raise ValueError("On-demand sandbox activity worker registration requires a task hub name.")
+
+    if not worker_profile_id or not worker_profile_id.strip():
+        raise ValueError("On-demand sandbox activity worker registration requires a worker profile ID.")
+
+    if max_activities_count <= 0:
+        raise ValueError("On-demand sandbox activity worker max activity count must be greater than zero.")
+
+    resolved_activity_names = resolve_activity_names(activity_names)
```
YunchuWang: This normalization is redundant in the worker path because `_configure_on_demand_sandbox_activity_filters()` already resolves and validates the activity list before registration messages are built. If this builder becomes private/internal, let's keep a single validation point and pass the resolved names through here directly.

durabletask-azuremanaged/durabletask/azuremanaged/preview/on_demand_sandbox/client.py
```diff
@@ -0,0 +1,382 @@
...
+            worker_profile_id: str) -> pb.RemoveOnDemandSandboxActivityDeclarationResult:
+        return self._stub.RemoveOnDemandSandboxActivityDeclaration(
+            pb.RemoveOnDemandSandboxActivityDeclarationRequest(worker_profile_id=worker_profile_id))
+
+    def connect_on_demand_sandbox_activity_worker(
+            self,
+            messages: Iterable[pb.OnDemandSandboxActivityWorkerMessage]
+    ) -> pb.OnDemandSandboxActivityWorkerSessionResult:
+        return self._stub.ConnectOnDemandSandboxActivityWorker(messages)
+
```
YunchuWang: Reusing this client for the worker stream is fine since it shares the same stub/channel/auth setup, but this should not be a public customer-facing method. It takes internal protobuf messages and is only used by `OnDemandSandboxWorker`, so let's make it private or move it behind a worker-specific private helper.

durabletask-azuremanaged/durabletask/azuremanaged/internal/sandbox_service_pb2.py
```diff
@@ -0,0 +1,53 @@
...
+# source: durabletask/azuremanaged/internal/sandbox_service.proto
+# Protobuf Python Version: 5.26.1
+"""Generated protocol buffer code."""
+from google.protobuf import descriptor as _descriptor
+from google.protobuf import descriptor_pool as _descriptor_pool
+from google.protobuf import symbol_database as _symbol_database
+from google.protobuf.internal import builder as _builder
+# @@protoc_insertion_point(imports)
+
+_sym_db = _symbol_database.Default()
```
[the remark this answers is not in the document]
berndverst: This is generated from protos. Do not modify or give feedback on this.

durabletask/internal/orchestrator_service_pb2.pyi
berndverst, line 1: I assume this changed because you regenerated using the latest commit hash.

durabletask/internal/PROTO_SOURCE_COMMIT_HASH
```diff
@@ -1 +1 @@
-e42905306605b7be6efb789dc489d27e27e22f71
\ No newline at end of file
+8c46a707cd543dd0ce3f031cb19255a11dde8ed4
```
berndverst, line 1: Only use commits that are merged into main branch please.

Double check this commit is correct once merged to main, then regenerate.

examples/serverless/README.md
```diff
@@ -0,0 +1,77 @@
+# DTS serverless activities sample
+
+This sample mirrors the .NET serverless sample with two customer-owned pieces:
+
+1. A **declarer app** (`main_app.py`) that declares which activity should run
+   serverlessly, starts the orchestration, and waits for the result.
+2. A **remote worker image** (`remote_worker.py` plus `Containerfile`) that
+   DTS starts in a sandbox to execute the declared activity.
```
berndverst: Just like the other example, please use durable task scheduler in the text - not DTS

examples/serverless/README.md
```diff
@@ -0,0 +1,77 @@
+# DTS serverless activities sample
```
berndverst: Durable task scheduler

berndverst:
Are serverless and sandboxes different features? Maybe combine to a single example?

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/__init__.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Sandbox preview APIs for Azure Managed Durable Task Scheduler.
```
berndverst: First time I heard "Azure Managed Durable Task Scheduler"

I thought we just say "Durable Task Scheduler" or "Azure Durable Task Scheduler" or something like that.

durabletask/worker.py
```diff
@@ -1217,62 +1217,71 @@ def _cancel_orchestrator(
         )
         self._logger.info(f"Cancelled orchestration task for invocation ID: {req.instanceId}")
 
+    def _durabletask_on_activity_execution_started(self, req: pb.ActivityRequest) -> None:
```
berndverst: Drop the `durabletask` -- can just be `_on_activity_execution_started`

durabletask/internal/orchestrator_service_pb2.pyi
YunchuWang: This is generated protobuf output. Per the current direction, I did not modify pb2/pb2.pyi files in this pass. The sandbox proto source hash is now synced to the merged durabletask-protobuf main commit.

durabletask/internal/PROTO_SOURCE_COMMIT_HASH
YunchuWang: The sandbox proto source hash is now pinned to merged durabletask-protobuf main commit 67a2f1d3b9ea8a210eb86addb636a5d8a6484dbc. I did not regenerate core pb2 files in this pass to avoid changing generated protobuf output.

durabletask-azuremanaged/durabletask/azuremanaged/internal/sandbox_service_pb2.py
```diff
@@ -0,0 +1,55 @@
...
```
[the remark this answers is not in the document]
YunchuWang: This file is generated by protoc. Leaving generated pb2 output untouched rather than hand-editing the unused _sym_db binding.

YunchuWang:
The current sample uses `examples/sandboxes` and the API surface is named `sandboxes` to match the merged Durable Task Scheduler sandbox proto. I kept this as a single sandbox activity sample rather than adding/combining another serverless sample.

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py
```diff
@@ -0,0 +1,248 @@
...
+            token_credential=resolved_token_credential,
+            secure_channel=resolved_secure_channel,
+            concurrency_options=concurrency_options)
+
+        self._sandbox_taskhub = resolved_taskhub
+        self._sandbox_worker_profile_id = _resolve_worker_profile_id()
+        self._sandbox_activities: list[SandboxActivity] = []
+        self._sandbox_max_activities = resolved_max_concurrent_activities
+        self._sandbox_provider = resolved_sandbox_provider
+        self._sandbox_dts_sandbox_identifier = os.getenv("DTS_SANDBOX_ID")
```
berndverst: `DTS_SANDBOX_ID` is read here without validation, and `worker_messages.build_sandbox_worker_start` then sends `(dts_sandbox_identifier or "").strip()`. Every other field on the start message (task hub, worker profile ID, max activities, ≥1 activity) is validated — this one isn't.

The server **requires** a non-empty sandbox identifier in the worker-start message and rejects a blank value with `INVALID_ARGUMENT`. It's also the key the Backend uses to correlate the live session with the sandbox resource (reaping) and to enforce a single active session per sandbox. Combined with the catch-all retry loop (L143), a missing/empty `DTS_SANDBOX_ID` becomes a silent, infinite reconnect loop where the worker never registers.

Recommend failing fast at construction, mirroring the other `_resolve_*` required-env checks (this also matches the .NET SDK, which normalizes the sandbox ID as required).

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker_messages.py
```diff
@@ -0,0 +1,62 @@
...
+    if not resolved_activities:
+        raise ValueError("Sandbox activity worker registration requires at least one registered activity.")
+
+    message = pb.SandboxActivityWorkerMessage(
+        start=pb.SandboxActivityWorkerStart(
+            task_hub=taskhub.strip(),
+            worker_profile_id=worker_profile_id.strip(),
+            max_activities_count=max_activities_count,
+            sandbox_provider=_parse_sandbox_provider(sandbox_provider),
+            dts_sandbox_identifier=(dts_sandbox_identifier or "").strip()))
```
berndverst: This emits an empty `dts_sandbox_identifier` when `DTS_SANDBOX_ID` is unset. The server rejects a blank identifier with `INVALID_ARGUMENT`, so this should be validated as required like the fields above it (see the detailed note on `worker.py` L69).

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py
```diff
@@ -0,0 +1,248 @@
...
+                    host_address=self._sandbox_host_address,
+                    taskhub=self._sandbox_taskhub,
+                    token_credential=self._sandbox_token_credential,
+                    secure_channel=self._sandbox_secure_channel)
+                try:
+                    client.connect_sandbox_activity_worker(self._registration_messages())
+                    retry_delay = 1.0
+                finally:
+                    client.close()
+            except Exception as ex:
```
berndverst, line 145: The registration loop treats every exception as transient and retries forever with backoff, logging only at `warning`. But several worker-start failures are **permanent** misconfigurations the server returns as `INVALID_ARGUMENT`/`FAILED_PRECONDITION`:

- blank `dts_sandbox_identifier`
- a `worker_profile_id` with no declared profile
- registered activities that don't exactly match the declared profile's set
- `max_activities_count` not equal to the declared profile's max concurrency

Today each becomes an invisible infinite loop (new channel every cycle) instead of a clear startup failure. .NET fails fast / stops on fatal registration errors. Suggest distinguishing permanent failures (at minimum gRPC `INVALID_ARGUMENT`) — surface at `error` and stop/back off hard — while still retrying genuinely transient cases. Note some `FAILED_PRECONDITION`s (e.g. reconnect grace, sandbox-not-ready) are transient, so match on the specific cause rather than the status code alone.

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py
```diff
@@ -0,0 +1,248 @@
...
+    client_id = os.getenv("DTS_UMI_CLIENT_ID", "")
+    if not client_id.strip():
+        raise ValueError(
+            "Sandbox worker requires DTS_UMI_CLIENT_ID to be injected when "
+            "DTS_AUTHENTICATION is ManagedIdentity.")
+
+    return ManagedIdentityCredential(client_id=client_id.strip())
+
+
+def _resolve_sandbox_provider() -> str:
```
berndverst: `DTS_SANDBOX_PROVIDER` is hard-required here (raises unless `Sandbox`/`AcaSessionPool`), but the server doesn't read `sandbox_provider` in the worker-start path — it's effectively informational. Meanwhile `DTS_SANDBOX_ID`, which the server *does* require, isn't validated (L69).

So the validation posture is inverted vs. what's actually enforced: strict on the field the server ignores, lax on the field the server requires. .NET tolerates a missing/unknown provider (→ `UNSPECIFIED`). Recommend relaxing this to match (default to `UNSPECIFIED`) and moving the strictness to `DTS_SANDBOX_ID`.

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker_profiles.py
```diff
@@ -0,0 +1,108 @@
...
+    entrypoint: list[str] = field(default_factory=list[str])
+    cmd: list[str] = field(default_factory=list[str])
+
+
+@dataclass
+class SandboxWorkerProfileOptions:
+    """Options for a decorated sandbox worker profile."""
+
+    @dataclass(frozen=True)
+    class Activity:
```
berndverst: This nested `Activity` type duplicates the public `SandboxActivity` (helpers) — they're structurally identical, and `resolve_activities` normalizes both into `SandboxActivity`. `add_activity(...)` produces `SandboxActivity` while `add_activities([...])` expects this nested `Activity`, so two activity records sit side by side in the public surface. Consider collapsing to the single exported `SandboxActivity` for both methods and dropping the nested type — one record type is easier to document for a preview API.

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/client.py
```diff
@@ -0,0 +1,55 @@
...
+        self._transport = SandboxActivitiesGrpcTransport(
+            host_address=host_address,
+            taskhub=taskhub,
+            token_credential=token_credential,
+            channel=channel,
+            secure_channel=secure_channel,
+            interceptors=interceptors,
+            channel_options=channel_options)
+
+    def close(self) -> None:
```
berndverst, line 41: `SandboxActivitiesClient` exposes `close()` but no `__enter__`/`__exit__`, unlike the base DTS clients, and the sample (`examples/sandboxes/main_app.py`) never calls `close()` — so the gRPC channel leaks there. Recommend adding context-manager support and using `with SandboxActivitiesClient(...) as client:` in the example.

berndverst (commented):
## Review: Azure Managed sandbox preview APIs

Overall this is a solid, well-tested preview that faithfully mirrors the .NET design (env-var contract, defaults, profile-builder validation, activity-only worker model). I cross-checked the wire contract and resource validation against both the canonical sandbox proto and the actual server/Backend behavior the worker registers with.

**Verified parity (looks good):**
- Sandbox proto is pinned to the same commit the server vendors, so the wire contract is aligned.
- Client-side CPU/memory validation matches the server constant-for-constant (250m–16000m in 250m steps; 2048 MiB/core; same accepted formats). Nice that this is validated at declare time so users get the same error locally that the server would return.
- Env var names/defaults match .NET (heartbeat 2s, retry 1s→30s full-jitter, max concurrent activities 100, CPU tiers, memory-per-core).

**Priority items** (inline comments below):
1. `DTS_SANDBOX_ID` is not validated although the server requires a non-empty value — see `worker.py` L69 / `worker_messages.py` L38.
2. The registration loop retries permanent misconfigurations forever and silently — see `worker.py` L143.
3. The `DTS_SANDBOX_PROVIDER` strictness is inverted relative to what the server actually enforces — see `worker.py` L219.

Together, (1)+(2) mean a worker started without `DTS_SANDBOX_ID` (or with a profile/activity/concurrency mismatch) enters an invisible infinite reconnect loop instead of failing fast.

**Minor / Pythonic notes:**
- Two activity record types in the public surface (nested `Activity` vs the exported `SandboxActivity`) — inline on `worker_profiles.py` L38.
- `SandboxActivitiesClient` has `close()` but no context manager, and the sample never closes it — inline on `client.py` L41.
- `enable_sandbox_activities()` raises when no profiles are registered; .NET silently no-ops. Worth deciding if the raise is intended for a preview.
- `resolve_activities` casefolds the activity **name** for its dedup key but not the **version**; note the server matches the declared vs. registered activity sets case-sensitively, so client-side case-folding of names could (in rare mixed-case cases) collapse entries the server treats as distinct.
- Tests reset the global `_worker_profiles` registry via per-test `try/finally` `pop(...)`. An autouse fixture that snapshots/restores the dict would be more robust against a test forgetting to clean up.
- The root CHANGELOG entry for the activity-execution hooks describes an internal underscore-prefixed mechanism; per the repo's changelog guidance, consider reframing by user impact or dropping it as internal-only.

**Two server couplings worth documenting** (not bugs, but they only surface today as server rejections retried forever):
- The activity names the worker registers (Python function names) must exactly, case-sensitively match the names declared in the profile (incl. version).
- The worker's resolved max activity count must equal the declared profile's max concurrency. This holds at runtime because the env is server-injected, but there's no client-side guard.

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py
YunchuWang: done

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py
YunchuWang: retry on certain codes for now, add todo for future granular precondition cases error handling

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker_messages.py
YunchuWang: done

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker_profiles.py
YunchuWang: done

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/client.py
YunchuWang: done

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py
YunchuWang: done

durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+def _is_retriable_registration_failure(ex: Exception) -> bool:
+    if isinstance(ex, grpc.RpcError):
+        status_code = ex.code()
+        if status_code in _RETRIABLE_REGISTRATION_STATUS_CODES:
+            return True
+        return False
+
+    return isinstance(ex, OSError)
```
berndverst, line 264: Follow-up on the registration retry classification (`_is_retriable_registration_failure` / `_RETRIABLE_REGISTRATION_STATUS_CODES`).

This correctly fixes the earlier silent infinite-retry loop, but I think the `FAILED_PRECONDITION` default is now backwards for the most common transient case.

On the server's worker-start path, `FAILED_PRECONDITION` is not uniformly permanent. The important one is **reconnect grace**: when a worker's stream drops and it reconnects, the server holds the prior worker session and rejects the new stream with `FAILED_PRECONDITION` (`...already has an active worker session`) until the previous session's last heartbeat ages past a reconnect-grace window. Because this loop opens a brand-new channel on each attempt (and the server mints a new stream-registration id per Start), a routine reconnect within that window hits this *every* time — and the current code classifies it as permanent and stops the worker for good. That turns a transient network blip into permanent sandbox death and defeats the server's reconnect-grace mechanism. A couple of other `FAILED_PRECONDITION` cases (sandbox group still provisioning, profile not yet propagated/declared) are likewise self-correcting without restarting the worker.

By contrast, the genuinely permanent, never-self-correcting failures are the `INVALID_ARGUMENT` ones (task-hub metadata mismatch, blank sandbox id, empty activity set). So the safer split is close to the inverse of what's here:

- **Permanent / stop:** `INVALID_ARGUMENT` (plus auth denials like `UNAUTHENTICATED` / `PERMISSION_DENIED`)
- **Retriable:** `FAILED_PRECONDITION` — ideally bounded with backoff over a window long enough to outlast the reconnect grace + declarer propagation, escalating the log level if it persists, rather than an immediate permanent stop.

The `TODO` to selectively retry `FAILED_PRECONDITION` once the server exposes structured, machine-readable reasons is a good future refinement — but until then the default arguably should *retry* `FAILED_PRECONDITION` rather than treat it as fatal, otherwise normal reconnects regress. (`test_sandbox_registration_does_not_retry_failed_precondition_without_structured_reasons` would need to flip accordingly.)

YunchuWang: keeping this as always retriable for now as discussed

berndverst (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ADDED
 
+- Added activity execution lifecycle hooks used by Azure Managed sandbox
+  workers to report active activity counts while work is running.
```
berndverst: Minor wording suggestion on this entry. It currently describes the hooks only by their effect; since these are new overridable extension points on the dispatch loop, it would be clearer to name the methods and say where they fire:

```markdown
- Added overridable activity-dispatch hooks `_on_activity_execution_started`
  and `_on_activity_execution_completed` on `TaskHubGrpcWorker`, invoked
  immediately before each activity runs and in a `finally` after it completes
  or fails. Subclasses can override these to observe in-flight activity
  execution (for example, to track the number of activities currently
  running).
```

durabletask-azuremanaged/CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
...
+
+- Updates base dependency to durabletask v1.6.0.
+- Added preview `durabletask.azuremanaged.preview.sandboxes` APIs for
+  declaring Durable Task Scheduler sandbox activities, registering
+  worker profiles, and running sandbox activity workers.
+- Added optional activity versions to sandbox worker profile declarations and
+  sandbox worker registration.
+- Required sandbox workers to fail fast when `DTS_SANDBOX_ID` is missing,
+  treated `DTS_SANDBOX_PROVIDER` as optional metadata, and added context
+  manager support to `SandboxActivitiesClient`.
```
berndverst: A few suggestions to tighten this Unreleased section into a single sandbox entry:

- These three bullets collapse into one. The "Added optional activity versions…" and "Required sandbox workers to fail fast…/optional provider/context manager" lines aren't corrections to a previously shipped release — the whole sandbox feature is landing in this same release, so they shouldn't read as fixes; they're just how the feature behaves on first release.
- Referring to the new classes as "APIs" reads a little oddly; describing what the feature lets you do is clearer.

Keeping the dependency bump as its own (unrelated) line, the sandbox bullets could become:

```markdown
## Unreleased

- Updates base dependency to durabletask v1.6.0.
- Added preview support for Durable Task Scheduler on-demand sandbox
  activities under `durabletask.azuremanaged.preview.sandboxes`. Applications
  can declare sandbox activities (with optional per-activity versions),
  register sandbox worker profiles, and run a sandbox activity worker that
  establishes a live worker session and dispatches only the activities it
  has registered.
```

berndverst (changes requested):
Please update the CHANGELOG :)

berndverst (approved):

PR #160, opened 2026-06-26
Add scheduled tasks feature and delayed entity signals

andystaples:
## Summary

Adds a **scheduled tasks** feature (`durabletask.scheduled`) to the Python SDK,
bringing it to parity with the `Microsoft.DurableTask.ScheduledTasks` package in
durabletask-dotnet. A schedule periodically starts a target orchestration based
on a configurable interval and time window. The feature is built entirely on top
of durable entities plus a small helper orchestrator — no new sidecar/protocol
requirements.

To make schedules work, this PR also adds **delayed (scheduled) entity signals**
across the entity, orchestration-context, and client `signal_entity` APIs, and
teaches the in-memory test backend to honor them.

### `durabletask.scheduled` package

- `configure_scheduled_tasks(worker)` — registers the schedule entity and
  operation orchestrator on a worker (parity with .NET's `UseScheduledTasks`).
- `ScheduledTaskClient` — `create_schedule`, `get_schedule`, `get_schedule_client`,
  `list_schedules`.
- `ScheduleClient` — per-schedule `create`, `describe`, `update`, `pause`,
  `resume`, `delete`.
- Public DTOs: `ScheduleCreationOptions`, `ScheduleUpdateOptions`,
  `ScheduleDescription`, `ScheduleQuery`, `ScheduleStatus`.
- Errors: `ScheduleNotFoundError`, `ScheduleInvalidTransitionError`,
  `ScheduleClientValidationError`.

Schedules support `interval` (≥ 1s), `start_at`, `end_at`, and
`start_immediately_if_late`, with a state machine (`Uninitialized` → `Active`
⇄ `Paused`) and an execution-token mechanism that cancels stale timer signals,
mirroring the .NET design.

### Delayed entity signals (prerequisite)

- Added an optional `signal_time` parameter to:
  - `EntityContext.signal_entity` / `DurableEntity.signal_entity`
  - `OrchestrationContext.signal_entity`
  - client `signal_entity` (sync and async)
- These map to the existing protobuf `scheduledTime` field. The public parameter
  name `signal_time` matches the .NET public API (`SignalEntityOptions.SignalTime`).
- The in-memory backend now defers entity operations whose `scheduledTime` is in
  the future (via a background timer thread), on both the client-signal and
  orchestration-emitted-signal paths.

## Tests

- **Unit**: schedule state machine, transition rules, config validation/diff,
  and state/options serialization round-trips.
- **In-memory E2E**: create/describe/run, pause/resume, update, list with
  prefix/status filters, delete; plus delayed-signal deferral via client and
  orchestrator.
- **Live DTS E2E** (`pytest.mark.dts`): scheduled-task lifecycle and delayed
  signals, verified against the DTS emulator.

## Docs & examples

- New `examples/scheduled_tasks.py`.
- `CHANGELOG.md` entries under Unreleased.

## Notes

- Placed in the core `durabletask` package (not a separate distribution as in
  .NET) for simpler installation and testing.
- Field names are Pythonic snake_case; class names stay aligned with .NET.
- The schedule entity name is lowercased to `schedule`, so list filtering uses
  the `@schedule@<prefix>` instance-ID prefix.


durabletask/testing/in_memory_backend.py
```diff
@@ -1760,7 +1785,30 @@ def _signal_entity_internal(self, entity_id: str, operation: str,
...
+        in-memory backend.
+        """
+        delay = max(0.0, (fire_at - datetime.now(timezone.utc)).total_seconds())
+
+        def fire() -> None:
+            time.sleep(delay)
+            with self._lock:
+                if self._shutdown_event.is_set():
+                    return
+                self._queue_entity_operation(entity_id, event)
```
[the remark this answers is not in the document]
andystaples: Fixed with a guard that checks whether the signal was sent in the same reset() generation

durabletask/internal/client_helpers.py
```diff
@@ -192,14 +192,16 @@ def build_terminate_req(
 def build_signal_entity_req(
         entity_instance_id: EntityInstanceId,
         operation_name: str,
-        input: Any | None = None) -> pb.SignalEntityRequest:
+        input: Any | None = None,
+        signal_time: datetime | None = None) -> pb.SignalEntityRequest:
     """Build a SignalEntityRequest for signaling an entity."""
+    scheduled_time = helpers.new_timestamp(signal_time) if signal_time is not None else None
```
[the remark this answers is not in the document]
andystaples: Incorrect - Protobuf already UTC-normalizes correctly, and every other callsite correctly trusts it. Introducing manual timezone change here would be inconsistent.

durabletask/entities/entity_context.py
```diff
@@ -94,15 +98,23 @@ def signal_entity(self, entity_instance_id: EntityInstanceId, operation: str, in
...
+        signal_time : datetime, optional
+            The time at which the signal should be delivered. If None, the signal is
+            delivered as soon as possible. Use this to schedule a future operation,
+            for example to have an entity wake itself up at a later time.
         """
         encoded_input: str | None = shared.to_json(input) if input is not None else None
+        scheduled_time: timestamp_pb2.Timestamp | None = None
+        if signal_time is not None:
+            scheduled_time = timestamp_pb2.Timestamp()
+            scheduled_time.FromDatetime(signal_time)
```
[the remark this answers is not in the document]
andystaples: Incorrect - Protobuf already UTC-normalizes correctly, and every other callsite correctly trusts it. Introducing manual timezone change here would be inconsistent.

durabletask/internal/helpers.py
```diff
@@ -255,11 +255,16 @@ def new_signal_entity_action(id: int,
...
                              operation: str,
                              encoded_input: str | None,
-                             request_id: str) -> pb.OrchestratorAction:
+                             request_id: str,
+                             scheduled_time: datetime | None = None) -> pb.OrchestratorAction:
+    scheduled_timestamp: timestamp_pb2.Timestamp | None = None
+    if scheduled_time is not None:
+        scheduled_timestamp = timestamp_pb2.Timestamp()
+        scheduled_timestamp.FromDatetime(scheduled_time)
     return pb.OrchestratorAction(id=id, sendEntityMessage=pb.SendEntityMessageAction(entityOperationSignaled=pb.EntityOperationSignaledEvent(
```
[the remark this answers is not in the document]
andystaples: Incorrect - Protobuf already UTC-normalizes correctly, and every other callsite correctly trusts it. Introducing manual timezone change here would be inconsistent.

durabletask/scheduled/client.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+        results: list[ScheduleDescription] = []
+        for metadata in self._client.get_all_entities(query):
+            state = _parse_state(metadata.get_state())
+            if state is None or state.schedule_configuration is None:
+                continue
+            if not self._matches_filter(state, filter):
+                continue
+            results.append(state.to_description())
+        return results
```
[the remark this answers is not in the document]
andystaples: Fixed

durabletask/scheduled/client.py
```diff
@@ -0,0 +1,158 @@
...
+        if filter is None:
+            return True
+        if filter.status is not None and state.status != filter.status:
+            return False
+        created_at = state.schedule_created_at
+        if filter.created_from is not None and not (created_at and created_at > filter.created_from):
+            return False
+        if filter.created_to is not None and not (created_at and created_at < filter.created_to):
+            return False
+        return True
```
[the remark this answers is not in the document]
andystaples: Fixed

durabletask/scheduled/registration.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from durabletask.worker import TaskHubGrpcWorker
+from durabletask.scheduled.orchestrator import \
+    execute_schedule_operation_orchestrator
```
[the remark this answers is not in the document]
andystaples: Fixed

durabletask/scheduled/registration.py
```diff
@@ -0,0 +1,20 @@
...
+    """Register the scheduled tasks entity and orchestrator with a worker.
+
+    Call this before starting the worker to enable scheduled tasks support.
+
+    Parameters
+    ----------
+    worker : TaskHubGrpcWorker
+        The worker to register the schedule entity and operation orchestrator with.
+    """
+    worker.add_entity(Schedule, ENTITY_NAME)
```
berndverst, line 20: `configure_scheduled_tasks` registers the entity + orchestrator but never advertises the scheduled-tasks worker capability. .NET's `UseScheduledTasks()` adds `WorkerCapability.ScheduledTasks` to the gRPC worker options, and the protobuf here already defines `WORKER_CAPABILITY_SCHEDULED_TASKS` — but it's referenced nowhere in Python source (`worker.py` only ever appends `LARGE_PAYLOADS`).

Recommend wiring this capability into the `GetWorkItemsRequest` capabilities list (mirroring the `_payload_store -> LARGE_PAYLOADS` pattern) so the backend can gate/observe scheduled-tasks support consistently with .NET — or confirm the DTS backend doesn't require it.

durabletask/scheduled/client.py
```diff
@@ -0,0 +1,142 @@
...
+    def _matches_filter(state: ScheduleState, schedule_query: ScheduleQuery | None) -> bool:
+        if schedule_query is None:
+            return True
+        if schedule_query.status is not None and state.status != schedule_query.status:
+            return False
+        created_at = state.schedule_created_at
+        if schedule_query.created_from is not None and not (created_at and created_at > schedule_query.created_from):
+            return False
+        if schedule_query.created_to is not None and not (created_at and created_at < schedule_query.created_to):
+            return False
```
berndverst, line 153: Two issues in the `created_from`/`created_to` comparison:

1. **Naive vs aware `datetime` crash.** `state.schedule_created_at` is timezone-aware (built from `datetime.now(timezone.utc)`), but `ScheduleQuery.created_from`/`created_to` are user-supplied and may be naive. `aware > naive` raises `TypeError: can't compare offset-naive and offset-aware datetimes`. The entity code already guards this with `_ensure_aware()`; the client filter should too.
2. **Exclusive bounds.** `>`/`<` drop a schedule whose `created_at` exactly equals the bound. Range filters are conventionally inclusive — consider `>=`/`<=`.

e.g. `_ensure_aware(created_at) >= _ensure_aware(query.created_from)`.

berndverst (commented):
## Review summary

Thanks @andystaples — this is a solid, well-tested feature that mirrors the .NET `ScheduledTasks` design closely: the `Schedule` entity state machine, the execution-token mechanism that cancels stale timer signals, and the per-operation orchestrator so client writes can `await` entity completion. Lint is clean and the scheduled + delayed-signal unit/in-memory-E2E tests pass locally (48 passed).

The recent commits are good improvements: making `ScheduleState.from_json` converter-free means state round-trips under any code path (not just the worker's threaded converter), and `signal_time -> scheduledTime` is now correctly threaded through both the client request builder and the orchestrator-action builder (these previously hardcoded `scheduledTime=None`).

A few things I'd like addressed or explicitly deferred (details inline):

**Recommend before merge**
- **Naive/aware `datetime` crash** in `ScheduleClient._matches_filter`: comparing an aware `schedule_created_at` against a possibly-naive `created_from`/`created_to` raises `TypeError`. Cheap to fix + test.
- **Missing `WORKER_CAPABILITY_SCHEDULED_TASKS` advertisement**: .NET advertises this capability; Python defines it in the protobuf but never sends it. Either wire it through `configure_scheduled_tasks` / `worker.py`, or confirm the backend doesn't gate on it.

**Reasonable follow-ups (these match current .NET behavior)**
- Typed schedule exceptions flattened to `RuntimeError` across the orchestration boundary.
- `list_schedules` applies status/created filters client-side, so pages may be underfilled.
- Exclusive `created_from`/`created_to` bounds (consider inclusive).
- Cross-SDK entity-name casing (`@schedule@` vs `@Schedule@`).
- No async client surface (the SDK already has an async client; .NET offers async).

Overall: **mergeable with minor fixes** — I'd gate on the datetime fix and a decision on the capability flag; the rest can be tracked as follow-ups. Nice work.

durabletask/scheduled/client.py
```diff
@@ -0,0 +1,142 @@
...
+            input=input,
+        )
+        instance_id = self._client.schedule_new_orchestration(
+            execute_schedule_operation_orchestrator, input=request)
+        state = self._client.wait_for_orchestration_completion(
+            instance_id, timeout=self._operation_timeout)
+        if state is None or state.runtime_status != OrchestrationStatus.COMPLETED:
+            failure = state.failure_details if state else None
+            message = failure.message if failure else "unknown error"
+            raise RuntimeError(
```
berndverst, line 53: Typed schedule exceptions are flattened to a generic `RuntimeError` here. The entity raises `ScheduleInvalidTransitionError` / `ScheduleNotFoundError`, but they cross the orchestration boundary as a `FailureDetails` string, so callers can't `except ScheduleInvalidTransitionError`. This matches the current .NET limitation, but mapping known failure types back to the typed exceptions (falling back to `RuntimeError`) would give a more Pythonic, future-proof client API.

durabletask/scheduled/client.py
```diff
@@ -0,0 +1,142 @@
...
+        prefix = schedule_query.schedule_id_prefix if schedule_query and schedule_query.schedule_id_prefix else ""
+        page_size = (schedule_query.page_size if schedule_query and schedule_query.page_size
+                     else ScheduleQuery.DEFAULT_PAGE_SIZE)
+        query = EntityQuery(
+            instance_id_starts_with=f"@{ENTITY_NAME}@{prefix}",
+            include_state=True,
+            page_size=page_size,
+        )
+        results: list[ScheduleDescription] = []
+        for metadata in self._client.get_all_entities(query):
```
berndverst, line 130: Status/created filters are applied client-side *after* fetching `page_size` entities, so a page can come back underfilled (or empty) even when more matches exist further down. This mirrors .NET (explicit TODO at `DefaultScheduledTaskClient.cs:86`), so it's acceptable — but the `page_size` contract is misleading. Recommend either documenting that filtered results may be sparse, or looping additional pages until `page_size` matches are collected.

durabletask/scheduled/models.py
```diff
@@ -0,0 +1,317 @@
...
+
+@dataclass
+class ScheduleQuery:
+    """Query parameters for filtering schedules."""
+
+    DEFAULT_PAGE_SIZE = 100
+
+    status: ScheduleStatus | None = None
+    schedule_id_prefix: str | None = None
+    created_from: datetime | None = None
```
berndverst, line 133: Defense-in-depth for the naive/aware comparison flagged in `client._matches_filter`: consider a `__post_init__` that coerces `created_from`/`created_to` to aware UTC, so a naive datetime supplied by a caller can never reach the comparison and crash.

durabletask/scheduled/schedule_entity.py
```diff
@@ -0,0 +1,243 @@
...
+
+from durabletask.entities import DurableEntity, EntityInstanceId
+from durabletask.scheduled import transitions
+from durabletask.scheduled.exceptions import ScheduleInvalidTransitionError
+from durabletask.scheduled.models import (ScheduleConfiguration,
+                                          ScheduleCreationOptions, ScheduleState,
+                                          ScheduleUpdateOptions)
+from durabletask.scheduled.schedule_status import ScheduleStatus
+
+ENTITY_NAME = "schedule"
```
berndverst, line 17: Interop note: Python lowercases the entity name to `schedule`, yielding instance IDs like `@schedule@<key>`, whereas .NET uses `nameof(Schedule)` -> `@Schedule@<key>`. Schedules created by the two SDKs against the same task hub won't see each other (different instance-ID prefixes), and `list_schedules` filtering on `@schedule@` won't match .NET-created entities. The choice is documented here — just flagging the cross-SDK seam to confirm it's intentional.

durabletask/scheduled/registration.py
andystaples: Done — `configure_scheduled_tasks` now advertises `WORKER_CAPABILITY_SCHEDULED_TASKS` via a new `TaskHubGrpcWorker.add_capability()` (mirroring the `_payload_store -> LARGE_PAYLOADS` pattern), so it's threaded into the `GetWorkItemsRequest` capabilities list just like .NET's `UseScheduledTasks()`.

One note for the record: I checked the DTS backend source and it currently only gates on `HistoryStreaming` (`ProtoUtils.cs` → `Capabilities.Contains(WorkerCapability.HistoryStreaming)`); nothing reads `ScheduledTasks` yet. So advertising it is presently inert against the backend, but it matches .NET and is forward-compatible if a future backend observes it.

durabletask/scheduled/models.py
andystaples: Done — `ScheduleQuery.__post_init__` now coerces `created_from`/`created_to` to aware UTC (via the shared `helpers.ensure_aware`), so a naive datetime supplied by a caller is normalized at construction and can never reach the `_matches_filter` comparison. The client filter also normalizes the stored timestamp defensively, so both sides of the comparison are guaranteed aware. Added unit tests for the normalization.

durabletask/scheduled/client.py
andystaples: Fixed the crash. `_matches_filter` now normalizes the stored `created_at` to aware UTC before comparing (via the shared `helpers.ensure_aware`), and `ScheduleQuery.__post_init__` coerces `created_from`/`created_to` to aware UTC so a naive bound can never reach the comparison (defense-in-depth, per your models.py comment too). Added unit tests for the naive-bound case.

On the bounds: I kept them **exclusive** (`>`/`<`) rather than switching to inclusive. I verified against the .NET source — `DefaultScheduledTaskClient.cs` uses `ScheduleCreatedAt > filter.CreatedFrom.Value` and `ScheduleCreatedAt < filter.CreatedTo.Value`, both strict — so exclusive matches .NET exactly. Added a comment noting the parity rationale.

durabletask/scheduled/schedule_entity.py
andystaples: Confirmed intentional. The Python framework lowercases all entity names at registration (`EntityInstanceId`/`_Registry.add_named_entity`), so `Schedule` → `@schedule@<key>` is unavoidable without special-casing this entity, and it's consistent with how every other Python entity behaves. The cross-SDK seam (`@schedule@` vs .NET's `@Schedule@`) is real but follows directly from that framework convention, and it's documented at the `ENTITY_NAME` definition. Leaving as-is.

durabletask/scheduled/client.py
andystaples: Agreed this is the more Pythonic direction, but I'd like to defer it — and after checking .NET, I think the right fix is actually a framework-level one rather than something local to this client.

Findings:
- The .NET ScheduledTasks client does **not** reconstruct typed exceptions either — `DefaultScheduleClient.cs` throws a generic `InvalidOperationException($"Failed to ...: {state.FailureDetails?.ErrorMessage}")` for every write op. So the current `RuntimeError` is at exact parity with .NET.
- .NET's clean mechanism for this is `TaskFailureDetails.IsCausedBy<T>()` (string-matches `ErrorType`, base-type aware). Python's `FailureDetails` exposes `error_type`/`message`/`stack_trace` but has **no `is_caused_by` equivalent** — that's the real gap, and it's general-purpose (benefits all orchestration callers, not just schedules).
- Even with that, faithfully re-raising `ScheduleInvalidTransitionError` loses its structured fields (`from_status`, etc.) across the wire — only the formatted message survives — so a converter-style reconstruction would need an alternate `from_failure(message)` constructor.

I've filed a follow-up issue to add `FailureDetails.is_caused_by(...)` to the core SDK [link to follow]. I'd rather land that framework primitive first, then optionally layer typed schedule exceptions on top, than special-case it here and exceed .NET parity in this PR.

durabletask/scheduled/client.py
andystaples: Documented rather than changed the behavior, to stay aligned with .NET (which has the explicit TODO at `DefaultScheduledTaskClient.cs:86`). `list_schedules` now carries a `> [!NOTE]` in its docstring explaining that the status/created filters are applied client-side per page, so an individual page may contain fewer than `page_size` matches (or none) even when more matching schedules exist. Happy to revisit the "loop until N matches" approach as a follow-up if you'd prefer the stricter contract.

durabletask/scheduled/client.py
andystaples: Filed the follow-up issue for the framework primitive: #162 (`Add FailureDetails.is_caused_by()` to mirror .NET's `TaskFailureDetails.IsCausedBy<T>()`).

andystaples: #162

berndverst (approved):

PR #161, opened 2026-06-26
Fix custom serialization gaps from #154

andystaples:
## Summary

PR #154 introduced type-aware custom object serialization but left several round-tripping gaps. This PR closes them — and fixes a couple introduced by #154

Builds on / follows up: #154.

### 1. `to_json()` hook precedence on the serialize side
The encoder checked the dataclass / `SimpleNamespace` branches **before** the `to_json()` hook, so those types could never use a custom `to_json()` — and a dataclass with a field that is not JSON-serializable on its own would raise even when it provided a `to_json()` hook to handle that field. The serialize side now prefers `to_json()`, mirroring the decode side which already prefers `from_json()`.

### 2. Nested `to_json()` honored (drop `dataclasses.asdict`)
Dataclasses were encoded with `dataclasses.asdict()`, which recursively flattens nested dataclasses to plain dicts (bypassing their `to_json()`) and deep-copies every leaf. Dataclasses are now encoded via a shallow field mapping so `json.dumps` recurses through the same hook, honoring nested `to_json()` and avoiding the deep copy.

### 3. `from_json` / type-directed recursion symmetry
`to_json` is recursive but reconstruction only recursed through `list` / `Optional` / `Union` / dataclass fields. It now also recurses into **`dict`/`Mapping` values** and **`tuple` elements**. For cases the built-in recursion can&#39;t cover, a `from_json` classmethod may now optionally accept the active converter — `from_json(cls, value, converter)` — and call `converter.coerce(...)` / `converter.deserialize(...)` to rebuild nested typed values. The single-argument form is unchanged.

Fixed-length `tuple[T1, T2, ...]` coercion now **fails fast**: a JSON array whose length does not match the tuple arity raises a `TypeError` rather than silently dropping trailing elements (or producing a short tuple) via `zip`.

### 4. Entity state: serialized-at-rest representation
`StateShim` eagerly deserialized the entity&#39;s wire state in its constructor, before the caller could supply a target type. It now holds the entity state in its **serialized JSON string form at all times** — the raw wire payload is stored verbatim, and a live value supplied via `set_state` is serialized immediately. This single, always-serialized representation has three benefits:

- **Deferred, type-directed deserialization.** `get_state(intended_type)` hands the raw string and the requested type to the converter together, so a custom converter can deserialize directly into the target type.
- **No double-encoding on persist.** An unmodified payload is handed straight back to the wire unchanged.
- **Earlier error isolation.** A value that can&#39;t be serialized raises at `set_state` time — inside the failing operation, which rolls back — instead of after the whole batch has run.

A redundant serialize→deserialize round-trip in the legacy entity event path was replaced with `converter.coerce(...)`.

> [!NOTE]
> **Behavior change:** because the held state is always the serialized form, `get_state()` now returns a freshly reconstructed value on every call rather than a reference to a single cached object. Mutating a value returned by `get_state()` in place no longer affects the persisted state — write it back with `set_state()` to persist. This is documented in the changelog under CHANGED.

### 5. `enum.Enum` round-tripping
Non-`int` enums failed to serialize (the decode side already reconstructed them). Enums now serialize to their underlying `.value` and reconstruct to the member when a target type is supplied — standalone, as dataclass fields, and inside `list` / `dict` / `tuple`. (`IntEnum` / `IntFlag` already serialized as integers.)

### 6. Forward-reference resolution (Python 3.10 parity)
On **Python 3.10**, `typing.get_type_hints()` does not deep-resolve forward references nested inside container args (e.g. the element type of `list["TreeNode"]` on a self-referential dataclass), leaving a bare string / `ForwardRef`. Type-directed reconstruction then skipped nested coercion and returned raw dicts — so a recursive / self-referential dataclass would not reconstruct its nested values on 3.10 (it worked on 3.11+). `_build_dataclass` now resolves forward references in each field hint against the dataclass&#39;s defining module before coercing, restoring 3.11+ behavior. Unresolvable names are left untouched (harmless fall back to parsed JSON), and the destination type is still entirely caller-supplied, so the security model is unchanged.

### 7. Module consolidation + deprecation
The internal `json_codec` module (added in #154, unreleased) is merged into `durabletask.serialization` and its functions are made private; the supported, stable surface is the pluggable `DataConverter`. `durabletask.internal.shared.to_json` / `from_json` are deprecated with a `DeprecationWarning` and continue to work for backwards compatibility.

## Tests

- New `tests/durabletask/test_data_converter_roundtrip.py` — comprehensive `JsonDataConverter` round-trip matrix (plain/nested/recursive dataclasses, `to_json`/`from_json` hooks incl. nested, custom classes, enums, containers, `Optional`/`Union`, primitives).
- New `tests/durabletask/test_shared.py` — deprecation-warning coverage for the legacy shim.
- Extended serialization + entity-executor tests for each fix: hook precedence, nested-hook recursion, dict/tuple recursion, fixed-length-tuple length-mismatch, serialized-at-rest `StateShim` (deferred deserialization, no-double-encode passthrough, eager-serialize-on-`set_state`, commit/rollback), converter-aware `from_json`, and version-independent forward-reference resolution.
- `flake8` and `pyright --strict` clean on changed files.

## Intentional limitations (documented via tests)

- Multi-member `Union[A, B]` is left as parsed JSON rather than guessing a member (security tradeoff; `Optional[T]` works).
- `datetime`, `Decimal`, `set`, `bytes`, and arbitrary plain objects still require a `to_json` hook or a custom `DataConverter`; they raise a clear `TypeError` naming the type.

## Compatibility

No new breaking changes vs 1.5.0 — all fixes are additive or warnings-only, except the entity `get_state()` reference-identity behavior change noted in §4 (documented in the changelog). Payloads produced by older SDK versions still deserialize (including into `SimpleNamespace` when no type is supplied). `durabletask.azuremanaged` (which subclasses the core worker/client) continues to work.


durabletask/serialization.py
```diff
@@ -134,3 +173,269 @@ def _log_coercion_fallback(target_type: type, error: Exception) -> None:
...
+        if not args:
+            return tuple(cast(list[Any], value))
+        # Homogeneous ``tuple[T, ...]``.
+        if len(args) == 2 and args[1] is Ellipsis:
+            return tuple(_coerce_to_type(item, args[0], converter) for item in cast(list[Any], value))
+        # Fixed-length ``tuple[T1, T2, ...]`` -- coerce element-wise by position.
+        return tuple(
+            _coerce_to_type(item, t, converter)
+            for item, t in zip(cast(list[Any], value), args)
+        )
```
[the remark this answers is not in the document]
andystaples: Fixed. Fixed-length `tuple[...]` coercion now fails fast: `_coerce_to_type` raises a `TypeError` when the JSON array length doesn't match the number of tuple elements, instead of letting `zip` silently drop trailing elements or produce a short tuple. Added `test_coerce_to_type_fixed_length_tuple_too_long_raises` and `test_coerce_to_type_fixed_length_tuple_too_short_raises` covering both directions.

durabletask/internal/entity_state_shim.py
```diff
@@ -35,31 +57,53 @@ def get_state(self, intended_type: None = None, default: Any = None) -> Any:
...
         # ``intended_type`` may be a typing generic (e.g. ``list[int]``) at
         # runtime, which is not a ``type`` instance, so the guard is required.
-        if (self._current_state is not None
-                and isinstance(intended_type, type)  # pyright: ignore[reportUnnecessaryIsInstance]
-                and not isinstance(coerced, intended_type)):
+        if (isinstance(intended_type, type)  # pyright: ignore[reportUnnecessaryIsInstance]
+                and not isinstance(result, intended_type)):
             raise TypeError(
                 f"Could not convert state of type '{type(self._current_state).__name__}' to '{intended_type.__name__}'"
             )
```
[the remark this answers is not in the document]
andystaples: Fixed. The error message now reports `type(result).__name__` instead of `type(self._current_state).__name__`, so it names the value that actually failed the coercion rather than always reporting `str`.

Note this code path was also reworked in a later commit: `StateShim` now holds state in its serialized string form at all times (rather than tracking an `is_serialized` flag), so `get_state` always defers to `converter.deserialize(raw, intended_type)`. `result` is the reconstructed value, which is what the message now reports.

andystaples:
## Follow-up changes (this round)

A few additions and refinements on top of the original PR, captured here to keep the summary from growing unbounded.

### 1. New example: pydantic-backed `DataConverter`

Added `examples/custom_data_converter/` — a self-contained subproject showing how to plug a third-party serialization/validation library into the SDK via a custom `DataConverter`:

- `PydanticDataConverter` serializes `pydantic.BaseModel` subclasses with `model_dump_json()` and reconstructs (with validation) via `model_validate_json()` / `model_validate()`, delegating everything else to the default `JsonDataConverter`.
- In-process tests against the in-memory backend prove the round-trip end-to-end (typed input arrives validated, typed `Receipt` read back via `state.get_output(...)`, constraint violations fail the orchestration) — no sidecar/emulator/Azure required.

### 2. New extension point: `DataConverter.can_reconstruct(target_type)`

Building the example surfaced a real gap: inbound type-discovery hard-coded the default converter's notion of what's reconstructable (dataclasses / `from_json`), so a custom converter could **serialize** a type but never get that type back on the **inbound** path — a pydantic input annotated `order: Order` arrived as a plain `dict`.

Fixed this properly rather than working around it:

- `DataConverter` now exposes an overridable `can_reconstruct(target_type)`. `type_discovery` consults the **active converter** instead of a hard-coded gate, and the worker threads its converter into all four discovery call sites (activity in/out, orchestrator in, entity in).
- On the abstract/concrete split: `can_reconstruct` is **not** abstract — it's an optional inbound-discovery hook, and forcing it on every converter would break converters used only for entity state / custom status (and minimal test doubles) that never touch type discovery. The base returns a conservative `False` (a converter claims nothing unless it opts in); the dataclass/`from_json` policy moved into `JsonDataConverter` where it belongs.
- The pydantic example overrides `can_reconstruct` to recognize models and defers the rest to its `JsonDataConverter` fallback.

### 3. Internal classes: `data_converter` is now required

`_RuntimeOrchestrationContext`, `_OrchestrationExecutor`, `_ActivityExecutor`, and `_EntityExecutor` took `data_converter: DataConverter | None = None` and silently fell back to `JsonDataConverter()`. Since these are private and always constructed by the worker with its already-resolved converter, the fallback could mask a mismatched-converter bug (it would "work" but with the wrong serialization). Made the parameter required and dropped the fallback so any missed wiring fails loudly at construction. (No public-API impact; unreleased.)

### Misc

- Renamed the new hook `is_reconstructable` → `can_reconstruct` (more idiomatic; unreleased, so no compat concern).
- Fixed a Python 3.10 reconstruction gap: forward references nested in container args (e.g. `list["TreeNode"]` on a self-referential dataclass) weren't deep-resolved by `get_type_hints` on 3.10, so nested values came back as raw dicts. Now resolved against the class's defining module, matching 3.11+.
- Fixed-length `tuple[...]` coercion now raises on a length mismatch instead of silently truncating via `zip`; `StateShim` error message now names the reconstructed value's type rather than always `str`.

All `tests/durabletask` unit tests + the new example tests pass; `flake8` clean and `pyright` clean on changed source.

examples/custom_data_converter/src/app.py
```diff
@@ -0,0 +1,90 @@
...
+    # The single line that wires in the third-party converter. The same
+    # converter is handed to the worker and the client below.
+    converter = PydanticDataConverter()
+
+    with DurableTaskSchedulerWorker(
+        host_address=endpoint,
+        secure_channel=secure_channel,
+        taskhub=taskhub_name,
+        token_credential=credential,
+        data_converter=converter,
```
berndverst, line 55: **This raises `TypeError` at runtime — and points to a larger gap.**

`DurableTaskSchedulerWorker` doesn't accept a `data_converter` parameter — it appears *nowhere* in `durabletask-azuremanaged`, its `__init__` has no `**kwargs`, and it doesn't forward anything like it to `super().__init__()`. So this call fails with `__init__() got an unexpected keyword argument 'data_converter'`.

The example's tests don't catch it because they bypass `app.py` and construct the core `TaskHubGrpcWorker`/`TaskHubGrpcClient` directly, and example tests aren't in CI (`testpaths = ["tests"]`).

The real fix is to add `data_converter: DataConverter | None = None` to `DurableTaskSchedulerWorker.__init__` (and both clients) and forward it to `super().__init__(...)`. That also makes the whole `DataConverter` feature usable from DTS, which it currently isn't. Happy to open an issue/PR for that.

examples/custom_data_converter/src/app.py
```diff
@@ -0,0 +1,90 @@
...
+        w.add_activity(calculate_total)
+        w.add_activity(charge_payment)
+        w.start()
+
+        c = DurableTaskSchedulerClient(
+            host_address=endpoint,
+            secure_channel=secure_channel,
+            taskhub=taskhub_name,
+            token_credential=credential,
+            data_converter=converter,
```
berndverst, line 67: Same problem as the worker above: `DurableTaskSchedulerClient` (and the async client) don't accept `data_converter`, so this raises `TypeError` until the azuremanaged client/worker forward it to the core base class.

examples/custom_data_converter/src/converter.py
```diff
@@ -0,0 +1,106 @@
...
+        if isinstance(value, BaseModel):
+            # ``model_dump_json`` honors pydantic field serializers, aliases,
+            # and custom types (e.g. ``datetime`` -> ISO 8601) automatically.
+            return value.model_dump_json()
+        return self._fallback.serialize(value)
+
+    def deserialize(self, data: str | None, target_type: type | None = None) -> Any:
+        if data is None or data == "":
+            return None
+        if _is_model_type(target_type):
```
berndverst, line 94: Nice "handle my types, delegate the rest" shape. One limitation worth documenting (or handling): this only intercepts when the model is the **top-level** `target_type`. For `list[Model]`, `dict[str, Model]`, or a dataclass field typed as a model, `_is_model_type(...)` is `False`, so it falls to the `JsonDataConverter` fallback, which reconstructs elements via `type_ctor(value)` -> `Model(dict)`. Pydantic models reject a positional dict, so e.g. `call_activity(..., return_type=list[OrderItem])` would raise `TypeError`, and inbound discovery of such an annotation silently passes raw dicts instead.

The example's own shapes avoid this (nested `list[OrderItem]` lives *inside* `Order`, handled by pydantic's own recursion), so it works — but since the README calls this "the recommended pattern for a real custom converter," it'd help to either note the top-level-generic limitation or recurse into generics that wrap a model.

durabletask/serialization.py
```diff
@@ -134,3 +198,356 @@ def _log_coercion_fallback(target_type: type, error: Exception) -> None:
...
+    > A plain instance method would have ``self`` consume the value and is
+    > unsupported regardless of the converter-arity detection below.
+    """
+    if converter is not None and _hook_accepts_converter(hook):
+        return hook(value, converter)
+    return hook(value)
+
+
+@functools.lru_cache(maxsize=2048)
+def _hook_accepts_converter(hook: Any) -> bool:
```
berndverst, line 419: The converter-arity detection keys off "2+ positional params." A `from_json` hook that happens to take an unrelated second parameter with a default — e.g. `from_json(cls, value, strict=False)` — would be misdetected as converter-aware and receive the `DataConverter` as that argument (`hook(value, converter)` binds `strict=<converter>`). Edge case, but worth a one-line doc note that the second `from_json` parameter is reserved for the converter (or matching on the parameter name).

durabletask/internal/type_discovery.py
```diff
@@ -5,54 +5,36 @@
...
+an explicit type.
 
 Discovery is intentionally conservative: it only returns an annotation when the
-target is a *reconstructable* custom type (a dataclass, a ``from_json()``-capable
-type, or an ``Optional`` / ``list`` wrapping one). Primitive and unknown
-annotations resolve to ``None`` so that existing payloads are passed through
-unchanged -- inbound type discovery never invokes an arbitrary constructor on
-untrusted data, and never alters the value for builtins.
+active :class:`~durabletask.serialization.DataConverter` reports it as
+*reconstructable* via :meth:`DataConverter.is_reconstructable`. The default
```
berndverst: `DataConverter` has no `is_reconstructable` method — it's `can_reconstruct`. (The old module-level `is_reconstructable` function was removed in this PR, so this is a stale reference.)
```suggestion
*reconstructable* via :meth:`DataConverter.can_reconstruct`. The default
```

berndverst (commented):
Reviewed at head `8db4510`. The seven documented fixes are real and correctly implemented, the `can_reconstruct` extension design is sound, and the CHANGELOG is honest and well-consolidated. Two things to flag — one that needs a fix before the feature is actually usable, plus a changelog suggestion — and a few smaller inline notes.

> [!NOTE]
> Baseline for this review is **v1.6.0** — both #154 and #161 landed after the v1.6.0 changelog entry and are still unreleased, so the whole `## Unreleased` block is the v1.6.0 -> next delta.

## Most important: the custom `DataConverter` isn't wired through the Azure-managed (DTS) worker/client

`data_converter` appears **nowhere** in `durabletask-azuremanaged`. `DurableTaskSchedulerWorker` / `DurableTaskSchedulerClient` (sync + async) don't accept it, have no `**kwargs`, and don't forward it to `super().__init__()`. Consequences:

- This PR's runnable example (`examples/custom_data_converter/src/app.py`) calls `DurableTaskSchedulerWorker(data_converter=...)` / `DurableTaskSchedulerClient(data_converter=...)` and will raise `TypeError` at runtime (inline comments below). The example's *tests* pass only because they bypass `app.py` and use the core in-memory `TaskHubGrpcWorker`/`TaskHubGrpcClient`; example tests aren't run in CI (`testpaths = ["tests"]`), so nothing catches this.
- More broadly, the headline feature of #154/#161 is currently **unreachable from DTS** — the package most production users run.

Fix is small: add `data_converter: DataConverter | None = None` to the three azuremanaged `__init__`s and forward it to `super().__init__(...)`. Then the example runs as written.

## Smaller items (see inline)
- **Entity `get_state()` reference-identity change** is a real, silent behavior change vs v1.6.0 for entities holding **mutable** state. No shipped example breaks (they all read-modify-`set_state` on immutable `int`), and it's documented under CHANGED — but consider making it a prominent migration note.
- **Pydantic example** only intercepts *top-level* model types; `list[Model]` / `dict[str, Model]` fall through to the fallback (converter.py note).
- **`_hook_accepts_converter`** arity heuristic can misfire on a `from_json(cls, value, unrelated=default)` hook (serialization.py note).
- **`type_discovery.py` docstring** references a non-existent `DataConverter.is_reconstructable` — should be `can_reconstruct` (suggestion inline).

Security and back-compat look solid: destination types are always caller-supplied (never read from the payload), `eval` in forward-ref resolution runs only on caller-authored annotation strings (never wire data), multi-member `Union` is left untouched rather than guessed, the `lru_cache`s are bounded, old `_AUTO_SERIALIZED` payloads still replay, and the deprecated `shared.to_json` / `shared.from_json` still work.

## Changelog suggestion

Since #154 and #161 both landed after v1.6.0 and are unreleased, the `## Unreleased` section should read as a single v1.6.0 -> next delta — which it now does well. Nothing from #154 is left dangling as a separate "fixed the unreleased thing" entry, and the falsy-state `FIXED` bullet is correctly a real fix vs v1.6.0 (v1.6.0 still used a truthiness check `entityState=...to_json(entity_state._current_state) if entity_state._current_state else None` that wrote `0`/`""`/`[]`/`{}` as `None`). Two suggestions:

**1. `durabletask-azuremanaged/CHANGELOG.md` is currently `N/A` under `## Unreleased`.** Once the worker/client forward `data_converter` (above), document the capability for the package that ships DTS support:

```markdown
## Unreleased

ADDED

- `DurableTaskSchedulerWorker`, `DurableTaskSchedulerClient`, and the async
  client now accept a `data_converter` argument and forward it to the base
  worker/client, so a custom `durabletask.serialization.DataConverter` (for
  example a pydantic-backed one) can be used with the Durable Task Scheduler.
```

**2. Optional:** make the entity `get_state()` `CHANGED` bullet state that it's a behavior change relative to the last release (v1.6.0) for entities that mutate state in place:

> `EntityContext.get_state()` / `DurableEntity.get_state()` now return a freshly reconstructed value on every call rather than a reference to a single cached object. **This changes v1.6.0 behavior:** mutating the returned value in place no longer affects persisted state — write it back with `set_state()`. State is also serialized eagerly at `set_state()` time, so a non-serializable value fails inside the operation (which rolls back) instead of after the batch has run.

Otherwise the folding is solid.

examples/custom_data_converter/src/app.py
andystaples: Fixed — both `DurableTaskSchedulerClient` and `AsyncDurableTaskSchedulerClient` now accept and forward `data_converter` to their core base clients, matching the worker.

examples/custom_data_converter/src/app.py
andystaples: Fixed — `DurableTaskSchedulerWorker` now accepts `data_converter: DataConverter | None = None` and forwards it to `super().__init__(...)`, so the `DataConverter` feature is usable from DTS. (Verified the converter instance is threaded through to the base worker.)

examples/custom_data_converter/src/converter.py
andystaples: Good catch — documented the limitation directly on `PydanticDataConverter` (a `> [!NOTE]` in the class docstring) and in the example README, and softened the "recommended pattern" wording to "a good starting point." The note explains that a model nested *inside another model* (like `Order.items`) round-trips via pydantic's own recursion (why the example works), but a model nested in a *top-level generic* the SDK rebuilds directly — `return_type=list[OrderItem]`, an input annotated `dict[str, OrderItem]` — is not intercepted and falls to the default codec, and that a production converter would recurse into such generics (e.g. via `pydantic.TypeAdapter`). I kept the example's surface minimal on purpose so it stays focused on the seam rather than turning into a full pydantic integration.

durabletask/serialization.py
andystaples: Fixed by matching on the parameter name rather than arity. `_hook_accepts_converter` now treats a hook as converter-aware only when its second positional parameter is named exactly `converter` (the reserved name), so `from_json(cls, value, strict=False)` is no longer misdetected — the converter would otherwise have been bound to `strict`. Updated the `from_json` NOTE to document that the second parameter name is reserved for the converter, and added `test_from_json_hook_unrelated_second_param_is_not_treated_as_converter` covering the `strict=False` case. (Both existing converter-aware hooks already name the parameter `converter`, so no behavior change for valid usage.)

durabletask/internal/type_discovery.py
andystaples: Fixed — updated the stale reference to `:meth:`DataConverter.can_reconstruct``.

examples/custom_data_converter/src/converter.py
andystaples: Note: This is an interesting follow-up: turning these kinds of shapes into first-class recursion scenarios within DataConverter/JsonDataConverter instead of a recursive external function. But, probably outside the scope for now.

berndverst (approved):

PR #173, opened 2026-07-14
Fix when_all to wait for all tasks before surfacing failures

andystaples:
## Summary

`task.when_all()` previously **failed fast**: as soon as one of its child tasks failed, the composite task was marked complete and the failure was surfaced immediately, even while sibling tasks were still running.

This does not match the .NET SDK ([durabletask-dotnet](https://github.com/microsoft/durabletask-dotnet)), which relies on the framework's `Task.WhenAll`. That contract is: the returned task completes only after **all** supplied tasks finish, and then surfaces the first fault.

This PR changes `WhenAllTask.on_child_completed` to record the first exception but only complete once every child task has completed — surfacing the first failure at that point (or building the ordered result list if none failed).

## Behavior change

An orchestration that fans out and hits a failure now waits for the remaining sibling tasks to finish before failing, instead of failing immediately. This is a deliberate change to match .NET semantics.

## Changes

- `durabletask/task.py` — `WhenAllTask.on_child_completed` no longer sets `_is_complete` on the first failure; it completes (with the first exception, if any) only after all children complete.
- `tests/durabletask/test_orchestration_executor.py` — updated `test_fan_in_with_single_failure` to assert the orchestration keeps running (zero actions) while tasks are outstanding, then fails once all children complete.
- `CHANGELOG.md` — added a `FIXED` entry.

## Testing

- `pytest tests/durabletask/test_orchestration_executor.py` — all pass
- `flake8` clean on changed files


durabletask/task.py
```diff
@@ -562,13 +562,20 @@ def on_child_completed(self, task: Task[Any]) -> None:
...
             raise ValueError('The task has already completed.')
         self._completed_tasks += 1
-        if task.is_failed and self._exception is None:
-            self._exception = task.get_exception()
-            self._is_complete = True
+        if task.is_failed:
+            self._failed_tasks += 1
+            if self._exception is None:
+                self._exception = task.get_exception()
         if self._completed_tasks == len(self._tasks):
```
berndverst, line 585: `_exception` is set as soon as the first child fails, while `_is_complete` remains false. That makes `is_failed` and `get_exception()` surface the failure before all child tasks have finished, which violates `when_all`'s wait-for-all behavior. Please retain the first exception separately and expose it only when the final child completes.

durabletask/task.py
```diff
@@ -552,6 +552,7 @@ def __init__(self, tasks: list[Task[T]]):
         super().__init__(cast(list[Task[Any]], tasks))
         self._completed_tasks = 0
         self._failed_tasks = 0
+        self._pending_exception: TaskFailedError | None = None
```
berndverst, line 567: The `_pending_exception` field is initialized after `super().__init__()`, but the base `CompositeTask` constructor invokes `on_child_completed()` for children that are already complete. If any such child is already failed, `on_child_completed()` reads `_pending_exception` before it exists and raises `AttributeError` instead of creating the `when_all` task. Please initialize this field before the base constructor runs, while preserving the completion state established for pre-completed children.

berndverst (approved):

PR #194, opened 2026-07-24
Add `FailureDetails.is_caused_by()` and fully-qualified error types

### LINKED ISSUE #162: Add `FailureDetails.is_caused_by()` to mirror .NET's `TaskFailureDetails.IsCausedBy<T>()`
## Summary

The .NET SDK exposes [`TaskFailureDetails.IsCausedBy<T>()` / `IsCausedBy(Type)`](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/TaskFailureDetails.cs), which lets a caller introspect an orchestration/activity failure and ask "was this caused by exception type `T`?" (with base-type support). The Python `FailureDetails` (`durabletask/task.py`) exposes `message`, `error_type`, and `stack_trace`, but has **no equivalent introspection helper** — callers must compare `failure_details.error_type` against a type name string by hand.

This is a general-purpose gap in the Python SDK's failure-introspection surface, not specific to any one feature.

## Background / motivation

This came up during the scheduled-tasks PR (#160). The schedule client surfaces entity-side failures (e.g. an invalid state transition) as a generic `RuntimeError` carrying the failure message — which matches the current .NET ScheduledTasks client behavior (it throws a generic `InvalidOperationException` with `FailureDetails.ErrorMessage`). A more Pythonic API would let callers do something like:

```python
try:
    schedule_client.pause()
except OrchestrationError as e:
    if e.failure_details.is_caused_by(ScheduleInvalidTransitionError):
        ...
```

But the building block for that — failure-type introspection — doesn't exist yet. Rather than special-case it in the scheduled-tasks client, the framework primitive should land first.

## Proposed API

Add to `durabletask.task.FailureDetails`:

```python
def is_caused_by(self, error_type: str | type) -> bool:
    """Return True if this failure's error_type matches `error_type`.

    Accepts either an exception type or its (qualified or unqualified) name.
    """
```

Design notes / open questions:
- .NET matches by attempting to **load** the type named in `ErrorType` and checking assignability (base-type aware). Python could do something analogous (import the type and use `issubclass`), or a simpler/safer name-based match. Worth deciding how much "load arbitrary type by name" behavior we want, given the security posture of the rest of the SDK.
- Python's `error_type` is the bare class name (e.g. `ScheduleInvalidTransitionError`), not namespace-qualified like .NET, so a name-based comparison is straightforward but won't disambiguate same-named types from different modules.
- Should this also consider `InnerFailure`/chained causes once nested failure details are exposed?

## Related

- PR #160 (scheduled tasks) — discussion thread on flattening typed schedule exceptions to `RuntimeError`.
- .NET reference: `src/Abstractions/TaskFailureDetails.cs`.

## Out of scope

Reconstructing fully-typed exceptions (e.g. re-raising `ScheduleInvalidTransitionError` with its structured fields) is a separate, larger concern — the structured fields don't survive the orchestration boundary today (only the formatted message does), so that would need an additional `from_failure(message)`-style constructor convention. This issue is only about the `is_caused_by` introspection primitive.

andystaples:
Fixes #162.

## What

Mirrors .NET's `TaskFailureDetails.IsCausedBy<T>()` and Java's `FailureDetails.isCausedBy(Class)` in the Python SDK.

- **New API — `FailureDetails.is_caused_by(error_type: str | type)`**
  - Passing an **exception type** performs a base-type-aware match: it returns `True` if the failure's type is `error_type` or any of its subclasses. It does this by walking the target type's already-imported subclass tree (visited-marked for diamond inheritance) and never imports a type by name, so it cannot execute arbitrary code. The safe trade-off is that a subclass whose module hasn't been imported yet won't be matched. Passing a non-exception type raises `TypeError`.
  - Passing a **string** matches by name only (no base-type awareness); the name may be qualified (`"module.ClassName"`) or unqualified (`"ClassName"`).

- **Fully-qualified error types (breaking)** — `FailureDetails.error_type`, and the `errorType` value sent over the wire, is now the fully-qualified `module.ClassName` (e.g. `builtins.ValueError`, `durabletask.task.TaskFailedError`) instead of the bare class name. This matches how the .NET (`GetType().ToString()`) and Java (`getClass().getName()`) SDKs already emit error types. A new `get_qualified_name()` helper is wired into `new_failure_details` and the in-memory backend. Protocol sentinels such as `"VersionMismatch"` are unchanged.

## Why fully-qualified names

The other SDKs store fully-qualified names, which is exactly how they avoid confusing same-named exceptions from different modules. Python previously stored only the bare `type(ex).__name__`, so this aligns the Python SDK with the rest and lets `is_caused_by()` disambiguate reliably.

## Compatibility notes

- **Breaking:** code comparing `error_type` against a bare name (e.g. `== "ValueError"`) must switch to the qualified name or, preferably, to `FailureDetails.is_caused_by()`.
- **Back-compat:** because this value is persisted and crosses the orchestration boundary, failures produced by older workers may still carry a bare name; `is_caused_by()` accepts both bare and qualified forms.
- `durabletask.azuremanaged` inherits the change; its changelog references the core notes.

## Design decision worth a look

Builtin exceptions keep their `builtins.` prefix (`ValueError` → `builtins.ValueError`), faithful to .NET/Java keeping `System.`/`java.lang.`. Easy to strip if reviewers prefer the shorter form.

## Testing

- Added `tests/durabletask/test_failure_details.py` (22 cases: exact/base-type/string/legacy-bare/FQN-disambiguation/error paths + end-to-end via `TaskFailedError.details`).
- Updated existing assertions that expected bare names.
- Core suite: 759 passed, 9 skipped. flake8 clean; pyright (strict) clean on changed source.

### DIFF
--- a/durabletask/internal/helpers.py
+++ b/durabletask/internal/helpers.py
@@ -122,6 +122,25 @@ def new_sub_orchestration_failed_event(event_id: int, ex: Exception) -> pb.Histo
     )
 
 
+def get_qualified_name(t: type) -> str:
+    """Return the fully-qualified ``module.qualname`` name of a type.
+
+    This mirrors the fully-qualified error-type names produced by the other
+    Durable Task SDKs (.NET ``Type.ToString()`` and Java ``Class.getName()``),
+    so that :attr:`FailureDetails.error_type` values are consistent across SDKs
+    and can be matched unambiguously by :meth:`FailureDetails.is_caused_by`.
+
+    Builtin exception types keep their ``builtins.`` prefix (e.g.
+    ``builtins.ValueError``), matching how .NET keeps ``System.`` and Java keeps
+    ``java.lang.``.
+    """
+    module = getattr(t, "__module__", None)
+    qualname = getattr(t, "__qualname__", None) or getattr(t, "__name__", None) or str(t)
+    if not module:
+        return qualname
+    return f"{module}.{qualname}"
+
+
 def new_failure_details(ex: Exception, _visited: set[int] | None = None) -> pb.TaskFailureDetails:
     if _visited is None:
         _visited = set()
@@ -130,7 +149,7 @@ def new_failure_details(ex: Exception, _visited: set[int] | None = None) -> pb.T
     if len(_visited) > 10 or (inner and id(inner) in _visited) or not isinstance(inner, Exception):
         inner = None
     return pb.TaskFailureDetails(
-        errorType=type(ex).__name__,
+        errorType=get_qualified_name(type(ex)),
         errorMessage=str(ex),
         stackTrace=wrappers_pb2.StringValue(value=''.join(traceback.format_tb(ex.__traceback__))),
         innerFailure=new_failure_details(inner, _visited) if inner else None
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+ADDED
+
+- Added `FailureDetails.is_caused_by()` to check whether a task failure was caused by a given exception type, mirroring .NET's `TaskFailureDetails.IsCausedBy<T>()` and Java's `FailureDetails.isCausedBy()`. Passing an exception type performs a base-type-aware match (a failure caused by a subclass matches its base type) against exception classes already imported in the process; passing a string matches by qualified or unqualified name.
+
+CHANGED
+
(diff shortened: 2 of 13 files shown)


durabletask/internal/helpers.py
```diff
@@ -130,7 +149,7 @@ def new_failure_details(ex: Exception, _visited: set[int] | None = None) -> pb.T
     if len(_visited) > 10 or (inner and id(inner) in _visited) or not isinstance(inner, Exception):
         inner = None
     return pb.TaskFailureDetails(
-        errorType=type(ex).__name__,
+        errorType=get_qualified_name(type(ex)),
```
berndverst, line 152: This changes a public, persisted wire value from a bare type name to a fully qualified name. Existing consumers that compare, route on, persist, or deserialize values such as `"ValueError"` will break; accepting legacy names only through the new `is_caused_by()` helper does not preserve those usages. Please retain the current `errorType` representation or defer this change to a major-version release.

berndverst: I think we can release in next minor release with a breaking change note instead.

andystaples: This was understood in the design, and I accepted the trade-off due to consistency with .NET/Java and string-matching on the exact error name already being an anti-pattern, but if you want to block the review on it, we can open a discussion. The problem is that errorType is a protobuf field, so we can't just add a new field for the fully qualified one

berndverst (approved):

Issue #195, opened 2026-07-24
Performance [P3]: Cache Azure managed SDK version lookup


berndverst:
### What is the issue?

Every synchronous and asynchronous Azure managed interceptor construction performs a distribution metadata lookup to determine the SDK version.

### What is the impact?

The lookup adds filesystem and package-metadata work to client and worker construction. It is a small cost per instance, but it appears on cold startup and is repeated when applications create multiple clients or workers.

### Details about the issue including code reference

Relevant code:

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/internal/durabletask_grpc_interceptor.py#L25-L42

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/internal/durabletask_grpc_interceptor.py#L83-L94

Both interceptor constructors call importlib.metadata.version for durabletask-azuremanaged and provide an unknown fallback on failure.

### A potential or proposed solution

Resolve the SDK version once in a module-level cached helper and reuse it for every interceptor. Preserve the current unknown fallback and user-agent format.

Issue #187, opened 2026-07-24
Performance [P2]: Defer synchronous Azure token acquisition


berndverst:
### What is the issue?

Constructing a synchronous Azure managed interceptor performs a potentially blocking credential request before any RPC is made.

### What is the impact?

Client and worker startup latency includes an authentication round trip even when the instance is never used. This especially affects short-lived tools, tests, and applications that construct clients eagerly. For normal long-lived clients, this moves rather than eliminates first-request latency.

### Details about the issue including code reference

Relevant code:

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/internal/access_token_manager.py#L17-L27

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/internal/durabletask_grpc_interceptor.py#L44-L50

`AccessTokenManager.__init__()` calls `credential.get_token()`. `DTSDefaultClientInterceptorImpl.__init__()` creates that manager and immediately calls `get_access_token()`.

### A potential or proposed solution

Defer token acquisition until the first intercepted RPC and retain cached refresh-before-expiry behavior. Consider an explicit warm-up option if fail-fast authentication during construction is an intended contract, because lazy acquisition changes when authentication failures surface.

### Cold-start relevance

This is the primary direct startup cost for synchronous Azure managed clients and workers: construction invokes the credential before the first RPC. With DefaultAzureCredential or ManagedIdentityCredential, that can include environment probing or a network token request. SandboxWorker passes its managed-identity credential through this same synchronous construction path.

Relevant code:

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/worker.py#L92-L117

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py#L43-L64

PR #198, opened 2026-07-27
Performance [P1]: Single-flight async Azure token refresh

### LINKED ISSUE #192: Performance [P1]: Single-flight async Azure token refresh
### What is the issue?

Concurrent asynchronous RPCs can independently refresh the same expired Azure access token.

### What is the impact?

When a token enters its refresh window, a burst of `grpc.aio` calls can trigger many credential requests at once. This adds latency, increases identity-provider load, and can contribute to credential throttling. The synchronous implementation already avoids this duplicate work.

Relevant code:

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/internal/access_token_manager.py#L73-L90

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/internal/durabletask_grpc_interceptor.py#L114-L126

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/internal/access_token_manager.py#L32-L37

`AsyncAccessTokenManager.get_access_token()` checks expiry and directly awaits `refresh_token()` without a lock or in-flight refresh task. `DTSAsyncDefaultClientInterceptorImpl._intercept_call()` awaits it for every RPC. In contrast, `AccessTokenManager.get_access_token()` uses double-checked locking.

Add an `asyncio.Lock` or shared in-flight refresh task. After entering the guard, re-check expiry before calling the credential. Keep the existing refresh window and ensure credential failures propagate to all waiting calls without deadlocking.

### Cold-start relevance

The asynchronous client constructor correctly defers token acquisition, so this does not delay process construction. However, a concurrent burst of first RPCs while no token is cached has the same duplicate-acquisition behavior as refresh after expiry. The proposed single-flight guard therefore improves cold first-request latency under concurrency.

Relevant code:

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/internal/access_token_manager.py#L63-L90

berndverst:
## Summary

`AsyncAccessTokenManager.get_access_token()` checked expiry and awaited
`refresh_token()` with no synchronization. Because
`DTSAsyncDefaultClientInterceptorImpl._intercept_call()` awaits it on **every**
RPC, a burst of concurrent `grpc.aio` calls arriving while the token was
missing or inside its renewal window would each independently hit the
credential — a thundering herd of Entra/credential requests, added latency, and
avoidable throttling pressure. The synchronous `AccessTokenManager` already
guards this with double-checked locking; the async path did not.

This change mirrors that design for the async manager:

- `get_access_token()` now re-checks token validity **after** acquiring a
  refresh guard, so only one coroutine calls the credential per refresh window
  and the rest reuse the result.
- The guard is an `asyncio.Lock` created lazily **per running event loop**
  (kept in a small map guarded by a `threading.Lock`). An `asyncio.Lock` binds
  itself to the loop it is first awaited on, so a single lock created in
  `__init__` would raise if the manager outlived a loop or was shared across
  loops. Locks belonging to closed loops are pruned so the map cannot grow
  without bound.
- The fast path is unchanged: when a cached token is still valid, the outer
  check short-circuits and no lock is touched, so there is no added per-RPC
  overhead.

Failures propagate unchanged — the guard is released via `async with`, so a
credential error surfaces to the caller and a later attempt can still succeed.

### Backwards compatibility

No public API changes. Method signatures, return values, the expiry/renewal
margin behavior, exception propagation, logging, and sync-path thread-safety
are all preserved. The only new members are private (`_get_refresh_lock`,
`_refresh_locks`, `_refresh_locks_guard`).

### Scope

Deliberately narrow, limited to this issue. It does not touch deferred sync
token acquisition (#187) or SDK version lookup caching (#195), which are being
addressed separately in these same files.

## Verification

- **New tests** in `tests/durabletask-azuremanaged/test_durabletask_grpc_interceptor.py`,
  modeled on the existing sync `test_concurrent_refresh_performs_single_refresh`:
  - 16 concurrent cold-start callers result in exactly **1** credential call.
  - 16 concurrent callers after expiry result in exactly **1** refresh.
  - A valid cached token is returned without contacting the credential.
  - A credential failure propagates to waiters and does not deadlock; a
    subsequent attempt succeeds.
  - The manager is reusable across two distinct event loops (guards against
    binding the lock to a dead loop).
- **Confirmed the tests catch the bug**: with the source change reverted, the
  two single-flight tests fail with `1 != 16` and `2 != 17` — exactly the
  duplicate acquisition described in the issue.
- **Provider unit tests**: `test_durabletask_grpc_interceptor.py` +
  `test_sandboxes_extension.py` — 47 passed (42 at baseline, +5 new).
- **Core unit tests**: `tests/durabletask` excluding `*_e2e.py` — 681 passed,
  7 skipped, identical to baseline.
- **flake8** clean on both changed Python files; **pyright** reports 0 errors
  on the changed source; **pymarkdown** introduces no new findings (the two
  MD013 hits on the changelog are pre-existing lines).
- `test_dts_batch_actions.py` fails identically with and without this change
  (11 failures on a clean baseline); it requires a live Durable Task Scheduler
  endpoint.

Fixes #192

### DIFF
--- a/durabletask-azuremanaged/CHANGELOG.md
+++ b/durabletask-azuremanaged/CHANGELOG.md
@@ -8,6 +8,9 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## Unreleased
 
 - `FailureDetails.error_type` now carries the fully-qualified type name (e.g. `durabletask.task.TaskFailedError`) instead of the bare class name, and the new `FailureDetails.is_caused_by()` helper is available (both inherited from durabletask). See the core `durabletask` changelog for details, including the breaking-change notes.
+- Improved async access token refresh concurrency handling to avoid duplicate
+  refresh operations under concurrent access, matching the existing sync
+  behavior.
 
 ## v1.8.0
 
--- a/durabletask-azuremanaged/durabletask/azuremanaged/internal/access_token_manager.py
+++ b/durabletask-azuremanaged/durabletask/azuremanaged/internal/access_token_manager.py
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
+import asyncio
 from datetime import datetime, timedelta, timezone
 from threading import Lock
 
@@ -70,9 +71,35 @@ class AsyncAccessTokenManager:
         self._token = None
         self.expiry_time = None
 
+        # An asyncio.Lock binds itself to the event loop it is first used on, and this
+        # manager may outlive a loop or be shared across loops. Locks are therefore
+        # created lazily per running loop, guarded by a plain threading lock because
+        # different loops may run on different threads.
+        self._refresh_locks: dict[asyncio.AbstractEventLoop, asyncio.Lock] = {}
+        self._refresh_locks_guard = Lock()
+
+    def _get_refresh_lock(self) -> asyncio.Lock:
+        loop = asyncio.get_running_loop()
+        with self._refresh_locks_guard:
+            lock = self._refresh_locks.get(loop)
+            if lock is None:
+                # Discard locks belonging to loops that are no longer usable so the
+                # mapping does not grow without bound.
+                for stale_loop in [
+                        existing for existing in self._refresh_locks if existing.is_closed()]:
+                    del self._refresh_locks[stale_loop]
+                lock = asyncio.Lock()
+                self._refresh_locks[loop] = lock
+            return lock
+
     async def get_access_token(self) -> AccessToken | None:
         if self._token is None or self.is_token_expired():
-            await self.refresh_token()
+            async with self._get_refresh_lock():
+                # Re-check under the lock: a concurrent caller may have already
+                # refreshed the token while this one was waiting, so only a single
+                # credential request is made per refresh window.
+                if self._token is None or self.is_token_expired():
+                    await self.refresh_token()
         return self._token
 
     def is_token_expired(self) -> bool:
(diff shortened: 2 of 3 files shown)


berndverst:
Thanks for the review. Addressing the two automated comments here, since one of them rests on a claim I was able to check directly.

## Blank line after the copyright header — declining

The suggestion was to add a blank line between the license header and the first import. I checked `main` before acting, and `access_token_manager.py` already has **no** blank line there:

```python
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
from datetime import datetime, timezone
```

This PR adds `import asyncio` as the first import and removes nothing, so it preserves the file's existing style rather than introducing a deviation. The repo's "header, then blank line" convention applies to **new** files; reformatting this one would be an unrelated style change to a file that was already like this.

Happy to fix it as a separate cleanup if you would prefer the convention applied retroactively across existing files, but it seems out of scope here.

## Comprehension readability — agreeing, with a correction to the reason

The comment on the list comprehension in `_get_refresh_lock` said the formatting "violates flake8." That part is not accurate — I ran `python -m flake8` against the changed files and it exits **0 with no output**, so there is no lint violation to fix.

The underlying readability point is fair, though, so it is being applied on its own merits: the comprehension is hoisted into a named local and then iterated. No behaviour change.

## Also in this update

This branch is being merged up to latest `main`, which now includes #199 (`0f316cc`). That PR rewrote `AccessTokenManager.__init__` to defer the initial token acquisition, touching the same constructor this PR modifies, so the merge keeps both: main's deferred first acquisition and this PR's single-flight async refresh. The interaction is worth a look on re-review, since the deferred *first* acquisition now also flows through the refresh lock that this PR made single-flight for the async path.

andystaples (approved):

PR #201, opened 2026-07-27
Performance [P3]: Serialize history events in one pass

### LINKED ISSUE #191: Performance [P3]: Serialize history events in one pass
### What is the issue?

History event export deep-copies the dataclass graph and then recursively walks it again to convert values for JSON.

### What is the impact?

Exporting large orchestration histories performs avoidable allocations and CPU work. This is limited to the history export feature but can be material for large archive jobs.

Relevant code:

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask/history.py#L44-L45

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask/history.py#L329-L339

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask/extensions/history_export/serialization.py#L42-L52

`HistoryEvent.to_dict()` calls `_to_serializable(asdict(self))`. `dataclasses.asdict()` recursively deep-copies nested dataclasses, lists, and dictionaries; `_to_serializable()` then walks lists and dictionaries again. History export invokes this path through `event_to_dict()`.

Implement a one-pass recursive serializer over dataclass fields, lists, dictionaries, and datetimes instead of calling `asdict()`. Preserve the existing output shape, datetime format, and handling of nested history dataclasses.

berndverst:
## Summary

`HistoryEvent.to_dict()` was implemented as `_to_serializable(asdict(self))`. `dataclasses.asdict()` recursively **deep-copies** the entire event graph into a throwaway intermediate structure, and `_to_serializable()` then walked that copy **again** to convert datetimes and rebuild containers. That is two full traversals plus a discarded allocation per event, multiplied across every event in a history export.

This replaces it with a single-pass recursive serializer that reads dataclass fields directly:

- `_to_serializable()` now handles dataclass instances itself (converting them to dicts in declaration order) instead of relying on `asdict()` to pre-flatten them.
- A fast path returns exactly-JSON-native values (`str`, `int`, `float`, `bool`, `None`) before any further checks — most event fields hit it.
- Field-name tuples are cached per dataclass type via `functools.lru_cache`, since history export walks many events of the same handful of types.

The dataclass check mirrors the private `dataclasses._is_dataclass_instance` predicate that `asdict()` used (`hasattr(type(v), '__dataclass_fields__')`), so dataclass *types* remain leaves and only instances recurse.

## Preserving the `asdict` contract

Review correctly caught that the first version of this change did more than optimize. Most of the compatibility behavior of the old pipeline lived inside `asdict`, not in the `_to_serializable` pass that followed it, and walking values in place silently dropped it. `asdict` rebuilds `list`/`tuple`/`dict` **subclasses** through their own constructors, recurses into mapping **keys**, special-cases namedtuples and `defaultdict`, and `copy.deepcopy`s every leaf.

The fix is to make the inline fast paths **exact-type** rather than `isinstance`, so they apply only where walking in place is provably identical to the old output, and to hand every other value back to the real `asdict`:

```python
@dataclass
class _LegacyBox:
    value: Any

def _asdict_only(value):
    return asdict(_LegacyBox(value))['value']
```

Boxing a single value in a throwaway dataclass re-enters the interpreter's own `asdict`, so subclass constructor protocols, key recursion, `defaultdict`, namedtuples and leaf deep-copying all behave exactly as before. Delegating rather than reimplementing matters because those internals have shifted between releases and this package supports 3.10 through 3.14 — a hand-rolled copy risks divergence that only shows up on one CI matrix leg.

That is not hypothetical, and this PR hit it. Until 3.12, `_asdict_inner` rebuilt *every* `dict` subclass with `type(obj)(<generator>)`, which `defaultdict` rejects:

```text
TypeError: first argument must be callable or None
```

3.12 added a dedicated branch that forwards `default_factory`. So the legacy pipeline **raises** for a `defaultdict` on 3.10 and 3.11 and **succeeds** from 3.12 on. Delegation reproduces both for free — the parity test asserting the walker fails the same way legacy does passes on every version. Had the branches been hand-rolled from a 3.13 reading of `_asdict_inner`, the walker would have happily serialized a value that the code it replaced could not, on exactly two of the five matrix legs.

Concretely, versus the first revision of this PR:

| input | first revision | now |
| --- | --- | --- |
| `datetime` subclass overriding `__deepcopy__` | converted from the original | converted from the copy, as before |
| `list` subclass | constructor skipped | constructor re-runs, as before |
| `dict` subclass | constructor skipped, keys reused | constructor re-runs, keys recursed |
| mapping key with `__deepcopy__` → `str` | left in place, `json.dumps` raises | deep-copied, serializes again |
| tuple subclass with iterator-only constructor | `TypeError` | rebuilt from a generator, as before |
| unrecognized mutable leaf | returned by reference | deep-copied |

**The tuple branch is removed entirely.** Normalizing datetimes inside tuples is a real bug — the legacy output for those payloads cannot be JSON encoded — but it is user-visible behavior, so it does not belong in a performance change and will be raised separately. Tuples now route through the compatibility path and come back exactly as before, quirk included. `TestTupleHandling::test_tuple_datetimes_remain_unencodable_like_legacy` pins the current behavior so that follow-up has to be deliberate rather than accidental.

### Cost

None. These paths are unreachable for SDK-produced events, so they were latent rather than live: the only `dict[str, Any]` field on any event is `HistoryStateEvent.orchestration_state`, populated exclusively by `_message_to_dict` → `json_format.MessageToDict(...)`, which yields JSON-native values only. Re-benchmarked against the shipped module after the change, the walker is **6.0x** faster than the original — the compatibility fallback never executes on the real corpus.

The field-name cache is now bounded (`lru_cache(maxsize=256)`) so that arbitrary user dataclass types nested under `orchestration_state` cannot grow it without limit. Measured over 200k lookups, best of 5: unbounded dict `0.0206s`, `lru_cache` `0.0204s`, no cache `0.1942s`. The bound is free; removing the cache is not an option at 9.4x slower.

## Output compatibility

Output is unchanged for every input that previously succeeded — same keys, same field ordering, same nesting, same `datetime.isoformat()` timestamps, same handling of `None`, enums, nested dataclasses, lists, and dicts — and inputs that previously raised still raise the same exception type. `to_dict()` still returns freshly built containers, and as with `asdict`, mutating any part of the result never touches the event.

No public API signature changed, nothing was renamed or removed, and no import path moved. `_to_serializable` keeps its name and signature.

Because this is now a pure internal performance change with no observable behavior difference, no changelog entry was added.

## Verification

- Added `tests/durabletask/test_history_serialization.py` (**157 tests**), an equivalence suite that pins `to_dict()` against a verbatim copy of the old two-pass implementation (`_to_serializable(asdict(event))`) used as a test oracle. It covers **all 28 `HistoryEvent` subclasses plus the base class** (39 sample instances: fully populated and minimally populated variants), with nested dataclasses, nested dicts/lists, naive and aware datetimes, empty containers, and a deeply nested `orchestration_state` payload. Each event is compared with both `==` and `repr()`, so key ordering and value types are pinned, not just equality. A guard test asserts every exported `HistoryEvent` subclass is covered.
- Added `TestLegacyCompatibility`, a parametrized corpus of 22 exotic values — container subclasses, namedtuples, `defaultdict`, iterator-only tuple constructors, custom mapping keys, sets, bytearrays and custom objects — asserting both that output matches the legacy pipeline and that failures match it too. The oracle is `asdict` applied to the **whole event**, never to an individual value, so it cannot become circular with the implementation's own delegation. Verified on 3.11 and 3.13 locally, either side of the `defaultdict` behavior change, in addition to CI's five versions.
- **RED → GREEN**: these tests were confirmed failing against the previous revision before the fix landed — **30 failures**, covering every point raised in review. Two fixtures had to be strengthened first: an idempotent subclass constructor and a `datetime` subclass overriding only `isoformat` both pass against the broken code, so they were replaced with a non-idempotent constructor and a `__deepcopy__` override respectively.
- Mutation-checked the equivalence assertions to confirm they have teeth: replacing the serializer with the identity function, and reversing field order, are each detected on all 39 sample events.
- Core unit tests: `903 passed, 7 skipped` (`pytest tests/durabletask -q --ignore-glob="*_e2e.py"`).
- Provider unit tests: `63 passed` (`test_sandboxes_extension.py`, `test_durabletask_grpc_interceptor.py`).
- `history_export` — the downstream consumer of `to_dict()` via `event_to_dict()` — `76 passed, 1 skipped`.
- `flake8 .` clean from the repository root, a superset of every per-directory target CI lints.
- `pyright` strict at the configured `pythonVersion` of 3.10 reports **0 errors** on both changed files.
- Benchmarked over the 39-event corpus (2000 iterations, best of 7), timing the **real `to_dict()`** rather than a transcription of it, so the number cannot drift from the code under review: original two-pass form `1.730s`, single-pass walker `0.287s` — **6.02x faster**. Output is asserted byte-identical event by event before any timing runs.

Fixes #191


durabletask/history.py
```diff
@@ -326,17 +330,67 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
+        return value
+    # Mirrors the private ``dataclasses._is_dataclass_instance`` check that
+    # ``asdict`` used: dataclass *types* are leaves, only instances recurse.
+    if hasattr(value_type, '__dataclass_fields__'):
+        return {
+            name: _to_serializable(getattr(value, name))
+            for name in _field_names(value_type)
+        }
     if isinstance(value, datetime):
         return value.isoformat()
```
andystaples, line 432: Previously, `asdict()` deep-copied datetime subclasses before `isoformat()` was called. Calling `isoformat()` directly on the original object can make a previously successful serialization throw or mutate the event when a subclass customizes `__deepcopy__` or `isoformat`. Please keep the fast path for exact `datetime` values while preserving legacy behavior for subclasses.

durabletask/history.py
```diff
@@ -326,17 +330,67 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
     if isinstance(value, list):
```
andystaples: `asdict()` reconstructed list subclasses through their concrete constructors before the old walker flattened them. Iterating the original directly bypasses those constructor semantics and can change the emitted values and wire shape. Please use an exact-type fast path and retain legacy-compatible handling for subclasses.

durabletask/history.py
```diff
@@ -326,17 +330,67 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
         return [_to_serializable(item) for item in cast(list[Any], value)]
+    if isinstance(value, tuple):
+        items = [_to_serializable(item) for item in cast(tuple[Any, ...], value)]
+        # Namedtuples take their fields as positional arguments rather than
+        # a single iterable, so rebuilding them needs the unpacked form.
+        if hasattr(value_type, '_fields'):
+            return value_type(*items)
+        return value_type(items)
```
andystaples: `asdict()` rebuilt tuple subclasses from a generator. This now passes a list, so an iterator-only tuple constructor that previously succeeded now raises `TypeError`. Please preserve the previous constructor protocol. Separately, normalizing datetimes inside tuples is observable behavior and should be handled in a dedicated bug-fix PR rather than this performance-only change.

durabletask/history.py
```diff
@@ -326,17 +330,67 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
     if isinstance(value, dict):
         return {
             key: _to_serializable(item)
             for key, item in cast(dict[Any, Any], value).items()
```
andystaples: The legacy `asdict()` pass reconstructed dict subclasses and recursively copied their keys. This now bypasses constructor behavior and reuses each original key. Besides changing wire values, a custom key whose `__deepcopy__` produced a JSON-native key previously serialized successfully; this path leaves the custom key in place and makes `json.dumps()` raise. Please preserve legacy-compatible dict and key handling.

andystaples (changes requested):
Requesting changes because this PR is described as a performance-only optimization, but the new walker changes serializer behavior for customer-supplied values.

Please keep this PR scoped to serializing history events in one pass while preserving the legacy `asdict()` + `_to_serializable()` contract for every input that previously succeeded. Exact built-in fast paths are fine, but subclasses, custom containers, and mapping keys need compatibility handling.

durabletask/history.py
```diff
@@ -326,7 +329,47 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
+    """Recursively convert *value* into a JSON-safe structure.
+
+    This walks dataclass instances directly instead of going through
+    ``dataclasses.asdict``, which would deep-copy the whole event graph
+    into a throwaway intermediate structure that then has to be walked a
+    second time. The output is identical to that two-pass form: nested
+    dataclasses become dicts in field order, datetimes become ISO 8601
+    strings, lists and dicts are rebuilt, and every other value is
+    passed through unchanged.
+    """
```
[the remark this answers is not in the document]
berndverst: Fixed. The fall-through now routes to `_legacy_compat()`, which re-enters the real `dataclasses.asdict` via a throwaway box dataclass, so leaves are deep-copied exactly as they were before rather than returned by reference.

Delegating instead of calling `copy.deepcopy` inline matters: `asdict` does more than copy at the leaf, and reproducing only the copy would still have diverged on container subclasses and mapping keys (see the other threads).

Covered by `TestToDictIsolation` and by `test_exotic_payload_matches_legacy_two_pass_output[custom-object]`.

durabletask/history.py
```diff
@@ -326,17 +330,67 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
+# strings, ints, or ``None``) before the ``isinstance`` chain below.
+_JSON_NATIVE_TYPES: frozenset[type[Any]] = frozenset({bool, float, int, str, type(None)})
+
+
+def _field_names(cls: type[Any]) -> tuple[str, ...]:
+    names = _FIELD_NAMES.get(cls)
+    if names is None:
+        names = tuple(field.name for field in fields(cast(Any, cls)))
+        _FIELD_NAMES[cls] = names
+    return names
```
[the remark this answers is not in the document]
berndverst: Good catch, and thanks for flagging it. Fixed in e969d30: `_field_names` is now a `functools.lru_cache(maxsize=256)` rather than an unbounded module-level dict, so an arbitrary number of user dataclass types nested under `orchestration_state` can no longer pin entries (or the classes they reference) indefinitely.

I measured the three options before choosing, 200k lookups, best of 5:

| variant | time |
| --- | --- |
| unbounded dict | 0.0206s |
| `lru_cache(256)` | 0.0204s |
| no cache | 0.1942s |

The bound is free, so keeping the fast path costs nothing. Dropping the cache entirely was not viable, it is 9.4x slower because `dataclasses.fields()` rebuilds a filtered tuple on every call.

durabletask/history.py
```diff
@@ -326,17 +330,67 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
+        }
     if isinstance(value, datetime):
         return value.isoformat()
     if isinstance(value, list):
         return [_to_serializable(item) for item in cast(list[Any], value)]
+    if isinstance(value, tuple):
+        items = [_to_serializable(item) for item in cast(tuple[Any, ...], value)]
+        # Namedtuples take their fields as positional arguments rather than
+        # a single iterable, so rebuilding them needs the unpacked form.
+        if hasattr(value_type, '_fields'):
```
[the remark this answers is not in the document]
berndverst: Agreed on the principle, and resolved by removing the cause rather than documenting it: the tuple branch is gone as of e969d30.

Tuples now fall through to the compatibility path and are rebuilt by `asdict` exactly as before, so a datetime inside a tuple is again left unconverted and the exported output is unchanged for every input. Since there is no longer any user-visible behavior change, a CHANGELOG entry would describe something that does not happen. This PR is back to being purely an internal performance change.

Normalizing datetimes inside tuples is still a real bug worth fixing, and it is genuinely user-visible, so it belongs in its own PR. `TestTupleHandling::test_tuple_datetimes_remain_unencodable_like_legacy` pins the current quirk so that follow-up is a deliberate, visible change rather than a silent one.

durabletask/history.py
```diff
@@ -326,17 +330,67 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
+            return value_type(*items)
+        return value_type(items)
     if isinstance(value, dict):
         return {
             key: _to_serializable(item)
             for key, item in cast(dict[Any, Any], value).items()
         }
```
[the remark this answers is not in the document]
berndverst: Fixed. The `dict` branch now runs non-native keys through `_asdict_only()`, which applies `asdict` recursion to the key and nothing else.

Worth noting why plain `copy.deepcopy` on the key was not sufficient: the legacy pipeline applied `asdict` recursion to keys but the conversion pass that followed it did not touch them, and those two differ. A frozen dataclass used as a key is the clearest case, `asdict` turns it into an unhashable dict and raises `TypeError`, whereas `deepcopy` would have silently succeeded and quietly widened the accepted input set. `_asdict_only` reproduces the original outcome, including the failure.

Covered by `test_mapping_key_is_deep_copied_like_asdict`, `test_mapping_key_is_not_converted_by_the_value_walker` and `test_dict_key_that_cannot_be_rebuilt_still_raises`.

durabletask/history.py
berndverst: Agreed, and fixed in e969d30. The datetime check is now an exact-type test (`type(value) is datetime`) rather than `isinstance`, so the fast path only applies where walking in place is provably identical to the old output. Subclasses fall through to `_legacy_compat()`, which re-enters the real `asdict` and therefore deep-copies before `isoformat()` is called, exactly as before.

You were right that this is observable and not merely theoretical. My first attempt at a regression test used a subclass that only overrode `isoformat`, and it passed against the broken code, because calling `isoformat` on the original and on a copy happen to agree. It only goes red with a subclass that customizes `__deepcopy__`, which is the case you called out. `test_datetime_subclass_deepcopy_substitution_is_honored` covers it, and it fails against the previous revision.

durabletask/history.py
berndverst: Agreed, and fixed. The list branch is now an exact-type check, so subclasses route to `_legacy_compat()` and are rebuilt through their own constructor before the conversion pass runs.

The same test-design trap applied here: a subclass whose constructor is idempotent, such as one that only dedupes, produces identical output either way and would have let the regression through. The fixture now appends a marker each time it runs, so the legacy output carries the marker twice, once from construction and once from the `asdict` rebuild, and skipping the constructor is directly visible. `test_list_subclass_constructor_still_runs` fails against the previous revision.

durabletask/history.py
berndverst: Agreed on both points, and both are done.

The tuple branch is deleted outright rather than repaired. Tuples now fall through to `_legacy_compat()`, which hands them to the real `asdict`, so the generator constructor protocol is preserved and an iterator-only constructor works again. `test_tuple_subclass_is_rebuilt_from_an_iterator` covers that and fails against the previous revision.

On the second point, you are right that normalizing datetimes inside tuples is observable behavior and does not belong in a performance change. It is out of this PR. `test_tuple_datetimes_remain_unencodable_like_legacy` pins the existing quirk, including the fact that the resulting payload cannot be JSON encoded, so the follow-up fix has to be an explicit, reviewed change rather than something that reappears by accident.

durabletask/history.py
berndverst: Agreed, and fixed. The dict branch is now an exact-type check, so subclasses route to `_legacy_compat()` and are rebuilt through their own constructor, and non-native keys go through `_asdict_only()` so key recursion and deep-copying behave as they did before.

Your `__deepcopy__` scenario is covered directly by `test_mapping_key_is_deep_copied_like_asdict`: a key whose `__deepcopy__` yields a string serialized fine before, raised on the previous revision, and works again now. As with the list case, the dict subclass fixture had to be made non-idempotent before it could detect the skipped constructor.

berndverst:
Thanks for the review — the framing was the useful part, and I agree with it. Most of the compatibility behavior of the old pipeline lived inside `asdict` rather than in the `_to_serializable` pass that followed it, and walking values in place quietly dropped it. That is a behavior change riding along with a performance change, which is not what this PR is for.

Rather than reimplement `_asdict_inner`'s branches, the fast paths are now **exact-type** checks and everything else is handed back to the real `asdict` through a throwaway box dataclass. That inherits whatever the running interpreter does, which matters given CI spans 3.10 through 3.14 and those internals have moved between releases.

The tuple branch is **removed outright** rather than repaired, per your point that it belongs in a dedicated bug-fix PR.

Two things worth flagging because they changed how I tested this:

- My first regression test for the `datetime` subclass case **passed against the broken code**. Overriding only `isoformat` is not enough, since calling it on the original and on a deep copy agree. It only goes red with a `__deepcopy__` override, which is the case you actually named.
- Same trap for the container subclasses: a constructor that only dedupes or uppercases is idempotent, so skipping it is invisible. The fixtures now append a marker each time they run, making the skipped constructor directly observable.

Both fixtures were strengthened before the fix landed. The new tests fail **30 times** against the previous revision and pass now.

`TestLegacyCompatibility` is the durable artifact you asked for: a parametrized corpus of exotic inputs asserting the walker matches `_to_serializable(asdict(...))`, including matching *failures*, so the contract is executable and the next optimization cannot silently break it. The oracle applies `asdict` to the whole event rather than to individual values, so it never becomes circular with the implementation's own delegation.

Re-benchmarked as requested: still **6.1x**, in fact marginally faster than before, since removing the tuple branch removed a check from the hot path. The compatibility work costs nothing on real payloads because those paths are unreachable — `orchestration_state` is fed exclusively by `MessageToDict`, which emits JSON-native types only.

CHANGELOG left alone: with the tuple branch gone there is no user-visible behavior change left to document.

berndverst:
Follow-up: the 3.10 and 3.11 legs went red on the previous push, and the cause is worth recording here because it is direct evidence for the design choice above.

`dataclasses.asdict` rebuilt **every** `dict` subclass with `type(obj)(<generator>)` until 3.12, and `defaultdict` rejects that:

```text
TypeError: first argument must be callable or None
```

3.12 added a dedicated branch forwarding `default_factory`. So the legacy pipeline **raises** for a `defaultdict` on 3.10 and 3.11 and **succeeds** from 3.12 on.

The implementation was not at fault and is unchanged. Because it delegates to the running interpreter's `asdict`, it reproduced both behaviors exactly — `test_exotic_payload_raises_the_same_way_as_legacy[defaultdict]` **passed on 3.10**, which is the assertion that actually matters. What failed was my strict-equality test, which assumed the legacy pipeline always yields a value to compare against. It now skips when legacy cannot serialize the payload at all and leaves that case to the parity test.

This is the failure mode that argued against hand-rolling `_asdict_inner`'s branches in the first place. A version-agnostic reimplementation written against a 3.13 reading of `_asdict_inner` would have serialized a `defaultdict` successfully on 3.10 and 3.11 — where the code it replaced raises — and that divergence would have been invisible on three of the five matrix legs. Added `test_defaultdict_follows_the_running_interpreter` so the next person to look at this finds the reason attached to the behavior.

Verified locally on **3.11** and **3.13**, either side of the change, in addition to CI: 156 passed / 1 skipped on 3.11, 157 passed on 3.13. Head is `e31bb48`, test-only.

berndverst:
Benchmark correction, since I quoted a specific figure above.

My earlier number came from a harness that re-implemented both walkers side by side. That is fine for isolating one branch, but it measures a transcription rather than the code under review, and the transcription had drifted — it still carried the tuple branch this PR no longer has.

Re-measured by importing `durabletask.history` and timing the real `HistoryEvent.to_dict()`, so the figure cannot diverge from what ships:

```text
corpus: 39 history events
legacy and head agree byte-for-byte on the real corpus

 legacy: best 1.730s  median 1.745s  (2000 iterations x 39 events)
   head: best 0.287s  median 0.290s  (2000 iterations x 39 events)

legacy -> head : 6.02x faster
```

**6.02x**, not the 6.1x I quoted. The compatibility fallback costs nothing measurable because it never executes on this corpus — every value reaches an exact-type fast path, which is the expected result given `orchestration_state` comes only from `MessageToDict`. The harness asserts byte-identical output event by event before it times anything, so a regression cannot be reported as a speedup.

PR description updated to the measured figures.

durabletask/history.py
```diff
@@ -326,19 +330,106 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
+        return value.isoformat()
+    if value_type is list:
+        return [_to_serializable(item) for item in cast(list[Any], value)]
+    if value_type is dict:
+        # ``asdict`` recursed into keys but the conversion pass that followed
+        # it did not, so keys get ``asdict`` semantics only. Native keys are
+        # returned as-is because ``asdict`` leaves those untouched too.
+        return {
+            (key if type(key) in _JSON_NATIVE_TYPES else _asdict_only(key)):
+                _to_serializable(item)
```
andystaples: This serializes each value before the mapping's deep-copied keys have been fully reconstructed, unlike the legacy two-phase pipeline. For example, if two custom keys both return `"same-key"` from `__deepcopy__`, legacy `asdict()` collapses the entries first and `_to_serializable` only visits the surviving value. This comprehension visits the overwritten value before that collision is known; if it contains a datetime subclass whose `isoformat()` throws, serialization now fails where legacy serialization succeeded.

Please preflight the graph and fall back around the complete event/container when any nested value needs legacy handling, so copying/reconstruction still completes before conversion begins. A regression test with colliding deep-copied keys and a throwing discarded value would pin this ordering requirement.

andystaples (changes requested):
The previous findings are addressed and the PR is substantially closer to a performance-only change. One compatibility gap remains: applying the legacy fallback independently to child values does not preserve the old pipeline's ordering (`asdict(event)` completes before `_to_serializable` begins).

This can turn a previously successful customer value into an exception, so I think it still needs to be resolved before approval. Please use the complete legacy pipeline for an event/container whenever its nested graph is not wholly eligible for the exact-type fast paths, and add coverage for interactions among multiple custom values rather than only isolated leaves.

durabletask/history.py
```diff
@@ -326,19 +330,106 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
 
+def _legacy_compat(value: Any) -> Any:
+    """Reproduce the original ``asdict`` + walk pipeline for one value."""
+    return _legacy_walk(_asdict_only(value))
+
+
+def _to_serializable(value: Any) -> Any:
+    """Recursively convert *value* into a JSON-safe structure.
+
+    This walks dataclass instances directly instead of going through
```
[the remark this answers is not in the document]
berndverst: Agreed, and fixed in 15c29eb.

The wording predates the compatibility rework and was accurate when written — at that point the walker converted everything it touched. Once tuples and other exotic shapes started routing back through `_legacy_compat` to preserve their original behavior, "JSON-safe" became a promise the function no longer keeps. A tuple holding a `datetime` is the clearest counterexample: it comes back with the `datetime` intact and `json.dumps` rejects it, exactly as before this PR.

The docstring now says which values are JSON-native and which are deliberately left alone:

> Values the SDK itself produces convert to JSON-native types. Arbitrary values can also arrive through `dict[str, Any]` fields, and those keep whatever the original pipeline did with them -- which for a few shapes, such as a tuple holding a `datetime`, is not JSON-encodable. That is preserved on purpose rather than fixed here; see `_legacy_compat`.

Worth being explicit that this is a docstring correction, not a behavior change: the non-encodable output for those shapes is what `main` produces today, and preserving it is the point of the compatibility path.

tests/durabletask/test_history_serialization.py
```diff
@@ -0,0 +1,797 @@
...
+
+
+class TestTupleHandling:
+    """Tuples keep their original ``asdict`` semantics.
+
+    An earlier revision of this change added a dedicated tuple branch that
+    normalized datetimes inside tuples. That was a behavior change rather
+    than a performance change, so it was removed: tuples now route to the
+    compatibility path and come back exactly as they did before. The
+    quirk that a datetime nested in a tuple is left unconverted is
```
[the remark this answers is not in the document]
berndverst: Already reconciled — in the direction of your first option, matching the preserved-quirk behavior.

That text was from an earlier revision. The tuple branch was removed after review feedback that normalizing datetimes inside tuples is user-visible behavior and does not belong in a performance-only change, and the description was rewritten at the same time. It currently reads:

> **The tuple branch is removed entirely.** Normalizing datetimes inside tuples is a real bug — the legacy output for those payloads cannot be JSON encoded — but it is user-visible behavior, so it does not belong in a performance change and will be raised separately. Tuples now route through the compatibility path and come back exactly as before, quirk included.

So description and implementation agree: tuples are unchanged from `main`. These tests are the intended guard rather than a contradiction — `test_tuple_datetimes_remain_unencodable_like_legacy` pins the quirk so the follow-up fix has to be a deliberate act rather than an accidental side effect of some later refactor.

durabletask/history.py
```diff
@@ -326,19 +330,106 @@ def _message_to_dict(msg: Message) -> dict[str, Any]:
...
+
+    The type checks below are deliberately exact rather than
+    ``isinstance``. Only the built-in types are handled inline, because
+    only for those is walking in place provably identical to what
+    ``asdict`` produced. Subclasses, tuples and every other value are
+    routed to :func:`_legacy_compat`, which re-enters the real ``asdict``
+    so their original semantics -- constructor round-trips, key
+    recursion and deep-copied leaves -- are preserved exactly.
+    """
+    value_type = cast('type[Any]', type(value))
```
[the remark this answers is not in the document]
berndverst: The premise does not hold, so the cast stays — but you are right that it deserved an explanation, and 15c29eb adds one. Two other casts nearby genuinely were redundant, and those are gone.

**`type(value)` is not `type[Any]` here.** Because `value: Any`, pyright infers `type[Unknown]`, and strict mode rejects it. Removing just this cast:

```
history.py:413:5 - error: Type of "value_type" is partially unknown
    Type of "value_type" is "type[Unknown]" (reportUnknownVariableType)
history.py:418:16 - error: Argument type is partially unknown (hasattr)
history.py:421:38 - error: Argument type is partially unknown (_field_names)
3 errors
```

CI runs pyright in strict mode over `durabletask/`, so this would not merge. A declared annotation (`value_type: type[Any] = type(value)`) does not help either — pyright narrows to the assigned type and still reports `type[Unknown]`.

**The string literal is deliberate, not stylistic.** An unquoted `type[Any]` is evaluated on every call, constructing a throwaway `types.GenericAlias` each time. Measured over 2,000,000 calls:

```text
bare type(v)              best 0.0899s
cast('type[Any]', ...)    best 0.1493s
cast(type[Any], ...)      best 0.2570s   <- 1.72x slower than quoted
```

In the hottest function in the module, that matters.

**Where you were right.** Chasing this found two casts that really were doing nothing — `cast(list[Any], value)` and `cast(dict[Any, Any], value)` in the same function. Those branches test `value_type`, not `value`, so `value` is never narrowed and iterating it directly is already clean. Both removed; pyright still reports 0 errors.

**One more thing this turned up.** `value.__class__` is both cheaper than `type(value)` (0.0821s vs 0.0905s) and pyright-clean with no cast at all, which makes it a tempting swap. It is also wrong: `__class__` can be overridden, so an object reporting `__class__` as `str` would take the JSON-native fast path and be returned by reference, where `asdict` — which uses `type(obj)`, verified against the stdlib source — deep-copied it. `TestLegacyCompatibility::test_dispatch_uses_type_not_the_overridable_class_attribute` pins that, and I confirmed it fails when the swap is made rather than assuming it would.

Verification: 160 serialization tests, 906 core, 63 provider, `flake8 .` clean, pyright strict 0 errors.

durabletask/history.py
berndverst: You are right, and thank you for catching it -- this was a real divergence and I reproduced it before fixing.

The part I had missed is that the legacy pipeline is **two-phase**: `asdict` rebuilds every key and collapses any that compare equal afterwards, and only *then* does `_to_serializable` walk the survivors. My single-pass walker fused those phases, converting each value as it iterated, so it visited entries that `asdict` had already discarded. Invisible unless converting a discarded value does something -- which is exactly the case you constructed.

Reproduction, two keys that deep-copy to the same string where the losing entry holds a `datetime` subclass with a throwing `isoformat`:

```
legacy : {'same-key': 'survivor'}
head   : ValueError: isoformat exploded
```

Fixed in 47e0f72. The dict fast path now scans keys first and hands the whole mapping to `_legacy_compat` if any key is non-native:

```python
if value_type is dict:
    for key in value:
        if type(key) not in _JSON_NATIVE_TYPES:
            return _legacy_compat(value)
    return {key: _to_serializable(item) for key, item in value.items()}
```

Two things worth calling out about the shape of that check:

- **A non-native key is a sufficient trigger.** A collision can only happen if `asdict` rebuilds at least one key -- native keys pass through untouched, and a mapping's keys are already distinct by construction. So I do not need to detect collisions, only the possibility of one, which is a cheap per-key type check.
- **The scan has to finish before any value is converted.** Detecting the non-native key mid-comprehension and bailing out would already have visited the earlier entries, which reintroduces the same bug for any mapping where the throwing entry comes first.

`_legacy_compat` applies both phases to the delegated subtree, so nesting composes: an inner mapping can take the legacy path while the outer stays on the fast path and the result is still exact.

Tests, in `TestLegacyCompatibility`:

- `test_colliding_keys_discard_their_value_before_conversion` -- your scenario, asserts the export equals the legacy output `{'same-key': 'survivor'}` and does not raise.
- `test_colliding_keys_keep_the_last_entry` -- pins *which* entry survives, so the ordering is not just "does not crash".
- Two new `_EXOTIC_PAYLOADS` entries, which run through both the equality oracle and the raises-the-same-way oracle.

I checked these actually pin the fix rather than passing by luck: reverting `history.py` to the previous commit takes 3 of the 166 to RED, and they go green again with the fix in place.

Benchmark is unchanged -- 6.2x, head best 0.283s vs 1.761s. Every mapping the SDK produces has `str` keys (`orchestration_state` comes from `MessageToDict`), so the scan is a type check per key on the fast path and the delegation never fires in practice.

andystaples (approved):

PR #203, opened 2026-07-27
Performance [P3]: Cache Azure managed SDK version lookup

### LINKED ISSUE #195: Performance [P3]: Cache Azure managed SDK version lookup

berndverst:
## Summary

Both the sync (`DTSDefaultClientInterceptorImpl`) and async
(`DTSAsyncDefaultClientInterceptorImpl`) interceptor constructors called
`importlib.metadata.version('durabletask-azuremanaged')` directly. That call
walks distribution metadata on disk, so the work was repeated on **every**
interceptor construction — i.e. on every client and worker construction.

This change resolves the version once in a module-level `functools.lru_cache`d
helper (`_get_sdk_version()`) and reuses it for every interceptor.

### Preserved exactly

- The `"unknown"` fallback when the version cannot be determined (the helper
  keeps the original broad `except Exception`, which still covers
  `importlib.metadata.PackageNotFoundError`).
- The user-agent string format, `durabletask-python/{version}`, verbatim.
- The metadata key names and their order:
  `taskhub`, `x-user-agent`, and `workerid` (sync worker case only).

No public API signature changes, no renames, no removed symbols, no changed
import paths. The only observable difference is the removed repeat filesystem
work.

> [!NOTE]
> No changelog entry was added. This is an internal performance change with no
> user-visible behavior difference, which the repo guidelines say should not be
> documented in the changelog.

## Verification

Ran against a session-local virtualenv with both packages installed in
editable mode.

**Effectiveness (the actual fix).** A probe patched the module-level `version`
symbol and counted calls across three interceptor constructions (sync client,
sync worker, async):

| Variant | `version()` calls |
| --- | --- |
| Before (uncached) | 3 |
| After (cached) | 1 |

**Tests.**

- New `TestSdkVersionCaching::test_version_resolved_once_across_interceptor_constructions`
  asserts the lookup happens exactly once across three constructions, and pins
  the full metadata list (keys **and** order) for all three cases.
- New `TestSdkVersionCaching::test_unknown_fallback_when_version_cannot_be_determined`
  asserts `PackageNotFoundError` still yields `durabletask-python/unknown` for
  both the sync and async interceptors.
- Both new tests clear the cache in `setUp`/`tearDown` so no patched value
  leaks into other tests.
- Existing `test_worker_includes_workerid_header` and
  `test_user_agent_metadata_passed_in_request` still pass unchanged.

**Suites.**

- Provider unit tests (`test_durabletask_grpc_interceptor.py`,
  `test_sandboxes_extension.py`, `test_azuremanaged_grpc_resiliency.py`):
  47 passed.
- Core unit tests (`tests/durabletask`, excluding `*_e2e.py`):
  681 passed, 7 skipped.
- `flake8` clean on both changed files.
- `pyright` (repo `pyrightconfig.json`, strict): 441 errors before and after,
  none in the changed file — unchanged from baseline.

`*_e2e.py` suites and `test_dts_activity_sequence.py` need a live Durable Task
Scheduler endpoint and were not run.

Fixes #195

### DIFF
--- a/durabletask-azuremanaged/durabletask/azuremanaged/internal/durabletask_grpc_interceptor.py
+++ b/durabletask-azuremanaged/durabletask/azuremanaged/internal/durabletask_grpc_interceptor.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+from functools import lru_cache
 from importlib.metadata import version
 
 import grpc
@@ -17,6 +18,22 @@ from durabletask.internal.grpc_interceptor import (
 )
 
 
+@lru_cache(maxsize=1)
+def _get_sdk_version() -> str:
+    """Return the installed version of the azuremanaged package.
+
+    Resolving the version walks distribution metadata on disk, so the result is
+    cached and shared by every interceptor instance instead of being recomputed
+    on each client or worker construction. Falls back to ``"unknown"`` when the
+    version cannot be determined.
+    """
+    try:
+        return version('durabletask-azuremanaged')
+    except Exception:
+        # Fallback if version cannot be determined
+        return "unknown"
+
+
 class DTSDefaultClientInterceptorImpl (DefaultClientInterceptorImpl):
     """The class implements a UnaryUnaryClientInterceptor, UnaryStreamClientInterceptor,
     StreamUnaryClientInterceptor and StreamStreamClientInterceptor from grpc to add an
@@ -27,13 +44,7 @@ class DTSDefaultClientInterceptorImpl (DefaultClientInterceptorImpl):
             token_credential: TokenCredential | None,
             taskhub_name: str,
             worker_id: str | None = None):
-        try:
-            # Get the version of the azuremanaged package
-            sdk_version = version('durabletask-azuremanaged')
-        except Exception:
-            # Fallback if version cannot be determined
-            sdk_version = "unknown"
-        user_agent = f"durabletask-python/{sdk_version}"
+        user_agent = f"durabletask-python/{_get_sdk_version()}"
         self._metadata = [
             ("taskhub", taskhub_name),
             ("x-user-agent", user_agent)]  # 'user-agent' is a reserved header; use 'x-user-agent'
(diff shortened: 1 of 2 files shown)


berndverst:
Thanks for the review. Declining the indentation change, because the stated reason turns out not to hold.

The automated comment flagged the hanging indentation (around L162-163 and L187-190) as something that "may fail flake8." I checked rather than assuming: `python -m flake8` on the changed files exits **0 with no output**. There is no lint violation.

The style in question is the standard PEP 8 form for a wrapped compound-statement header — the continuation lines are indented further than the block body specifically so the two are visually distinguishable:

```
with ...          <- column 8   (statement)
        ...       <- column 16  (continuation of the header)
    body          <- column 12  (block body)
```

PEP 8 calls this out explicitly as the recommended way to resolve the ambiguity, so changing it would move away from the convention rather than toward it.

Separately, this branch is being updated to merge latest `main`, which had made the PR conflicting — `main` added ~94 lines to `test_durabletask_grpc_interceptor.py` and modified `durabletask_grpc_interceptor.py`, both files this PR touches. The resolution keeps both sides: main's new coverage and this PR's cached SDK-version lookup.

andystaples (approved):

Issue #196, opened 2026-07-24
Performance [P2]: Avoid eager worker imports from Azure managed package


berndverst:
### What is the issue?

Importing the Azure managed package eagerly imports the full core worker implementation just to re-export worker-related types.

### What is the impact?

Client-only Azure managed applications pay to import gRPC, protobuf, entities, serialization, and worker code during Python process startup. This can materially increase cold-start latency in serverless and short-lived processes.

### Details about the issue including code reference

Relevant code:

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask/__init__.py#L6-L15

The shared `durabletask` package initializer imports `GrpcChannelOptions`, payload-store types, and several symbols from `durabletask.worker`. Because the Azure managed package lives under the same `durabletask` namespace, importing anything from it (for example `durabletask.azuremanaged.client`) first initializes `durabletask/__init__.py` and therefore loads the entire worker dependency graph, even when an application only needs a scheduler client.

### A potential or proposed solution

Move lightweight shared types into a smaller module, or use lazy module exports through module `__getattr__` with `TYPE_CHECKING` imports. Preserve current public import paths, especially `from durabletask import ConcurrencyOptions` and the other re-exported symbols, and keep `__all__` intact so `dir()` and star-imports behave as before.

PR #208, opened 2026-07-27
Performance [P3]: Lazily load sandbox runtime API exports

### LINKED ISSUE #197: Performance [P3]: Lazily load sandbox runtime API exports
### What is the issue?

Importing the sandbox preview package eagerly imports both its client and worker runtime APIs.

### What is the impact?

Callers that only declare profiles or use a subset of the sandbox API load the worker graph, Azure Identity, and gRPC-related dependencies during process startup. This adds avoidable cold-start work for sandbox control-plane and configuration workloads.

Relevant code:

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/__init__.py#L18-L23

https://github.com/microsoft/durabletask-python/blob/55d8e0b413eb5dd75afa64f30c54fe300bf4ea21/durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/worker.py#L1-L27

The package initializer imports SandboxActivitiesClient and SandboxWorker. Importing SandboxWorker loads the sandbox runtime and ManagedIdentityCredential even when that runtime is not used.

Use lazy package exports through module __getattr__ and TYPE_CHECKING imports, or split profile-definition APIs into a lightweight module. Preserve the documented public names and import behavior.

berndverst:
## Summary

Importing `durabletask.azuremanaged.preview.sandboxes` eagerly imported both the sandbox
client and the sandbox worker. The worker module pulls in `ManagedIdentityCredential` from
`azure-identity` plus the core worker graph, so merely touching the sandboxes namespace — even
to declare a worker profile or reference a type — paid that entire cost at process startup.

This replaces the eager re-exports in the package initializer with PEP 562 lazy exports: a
module-level `__getattr__` resolves each public name from its defining submodule on first
access and caches it in the module globals. `TYPE_CHECKING` imports keep type checkers, IDEs,
and `pyright --strict` fully aware of the names.

Backwards compatibility is preserved exactly:

- Every exported name and its import path is unchanged.
- `__all__` is byte-for-byte identical, so `from ... import *` still exports the same six names.
- `dir()` still lists the public names (via a module-level `__dir__`).
- Submodules (`client`, `helpers`, `worker`, `worker_profiles`, ...) remain reachable as package
  attributes, matching the attribute binding the eager imports used to produce as a side effect.
- Unknown attributes still raise a normal `AttributeError`.

## Before / after

Measured on the same machine with `durabletask` and `durabletask.azuremanaged` installed
editable, median of 9 runs each, `HEAD~1` vs `HEAD`:

| Metric | Before (eager) | After (lazy) | Change |
| --- | --- | --- | --- |
| `-X importtime` cumulative for `durabletask.azuremanaged.preview.sandboxes` | 965.7 ms | 376.0 ms | **-61.1%** |
| Wall clock, full `python -c "import ...sandboxes"` process | 1144.1 ms | 532.5 ms | **-53.5%** |
| `len(sys.modules)` after import | 548 | 296 | **-252 modules** |

Raw `-X importtime` tail lines:

```text
before: import time:       852 |    965731 | durabletask.azuremanaged.preview.sandboxes
after:  import time:       845 |    375998 | durabletask.azuremanaged.preview.sandboxes
```

`azure.identity` alone accounts for ~890 ms of cumulative import time on this machine, and it is
no longer loaded unless `SandboxWorker` is actually used:

```text
before: worker loaded True  | azure.identity loaded True
after:  worker loaded False | azure.identity loaded False
```

### Interaction with #196

#196 makes the analogous lazy-export change to the core `durabletask/__init__.py`. Because the
sandboxes package is reached *through* `durabletask`, there are two independent doors to the
worker graph, and closing only one leaves the other open. Measured here as a full 2x2 on
`import durabletask.azuremanaged.preview.sandboxes` — module counts are deterministic, times are
the median of 7 runs of the summed `-X importtime` self-times (which include the ~87 ms
interpreter floor):

| Configuration | Modules | Median import time |
| --- | --- | --- |
| Neither (`main`) | 548 | 1079.9 ms |
| #196 alone | 548 — no change | 1037.6 ms |
| **#197 alone (this PR)** | **296** | **452.7 ms** |
| Both | 84 | 92.6 ms |

Two honest conclusions from that table:

- **#196 alone does nothing for this import path** (548 -> 548). The eager sandbox worker import
  removed by this PR was pulling the core worker graph in directly, bypassing the core
  initializer's laziness entirely. This PR is the load-bearing change for the sandboxes path.
- The 296 modules remaining after this PR are `grpc`/`protobuf` and friends, still pulled in by
  the parent `durabletask` package. That is #196's scope and is deliberately untouched here.
  With both changes landed the import drops to 84 modules — only 6 above the bare-interpreter
  floor of 78, i.e. from 470 modules of overhead down to 6.

The two PRs are therefore complementary rather than additive, and they touch disjoint files
(verified: no path overlap between the branches). These figures were reproduced independently in
this worktree; the module counts match the #196 author's measurements exactly.

## Verification

- **New tests** in `tests/durabletask-azuremanaged/test_sandboxes_extension.py`. Because that
  module imports the sandbox worker at module scope, the lazy-import assertions run in a
  subprocess:
  - `test_sandbox_package_defers_worker_runtime_imports` — asserts `...sandboxes.worker`,
    `...sandboxes.client`, `...sandboxes.transport`, and `azure.identity` are absent from
    `sys.modules` after importing the package, then asserts `__all__`, `dir()`, that every
    exported name resolves to the same object as a direct import, and that submodule attributes
    still work.
  - `test_sandbox_package_star_import_matches_all` — asserts `from ... import *` yields exactly
    the `__all__` names and does not leak the lazy-import plumbing.
  - `test_sandbox_package_raises_attribute_error_for_unknown_attributes` — asserts a proper
    `AttributeError` and correct `hasattr` behavior.
  - Confirmed these fail against the pre-change initializer (the lazy-import probe reports
    `client`, `transport`, `worker`, and `azure.identity` as eagerly imported).
- **Provider unit tests**: `test_sandboxes_extension.py` + `test_durabletask_grpc_interceptor.py`
  — 45 passed (42 before this change, plus the 3 new tests).
- **Core unit tests**: `pytest tests/durabletask --ignore-glob="*_e2e.py"` — 681 passed,
  7 skipped, no regressions against the recorded baseline.
- **flake8**: clean on both changed Python files. The `TYPE_CHECKING` imports are resolved
  properly rather than suppressed — pyflakes treats names listed in `__all__` as used, so no
  `noqa` was needed.
- **pyright** (repo `pyrightconfig.json`, strict): zero errors in the changed file and zero
  anywhere under `durabletask-azuremanaged/durabletask`.
- **pymarkdownlnt**: the changelog addition introduces no new findings.
- **autopep8**: `--diff` is empty for both changed files.

Fixes #197

### DIFF
--- a/durabletask-azuremanaged/CHANGELOG.md
+++ b/durabletask-azuremanaged/CHANGELOG.md
@@ -7,6 +7,10 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+- Importing `durabletask.azuremanaged.preview.sandboxes` no longer loads the sandbox worker
+  runtime or `azure-identity` up front. The package's public names are now resolved on first
+  use, which roughly halves import cost for callers that only declare sandbox worker profiles
+  or use the sandbox client. Exported names, `__all__`, and import paths are unchanged.
 - `FailureDetails.error_type` now carries the fully-qualified type name (e.g. `durabletask.task.TaskFailedError`) instead of the bare class name, and the new `FailureDetails.is_caused_by()` helper is available (both inherited from durabletask). See the core `durabletask` changelog for details, including the breaking-change notes.
 
 ## v1.8.0
--- a/durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/__init__.py
+++ b/durabletask-azuremanaged/durabletask/azuremanaged/preview/sandboxes/__init__.py
@@ -13,14 +13,22 @@ Usage::
         SandboxWorker,
         SandboxActivitiesClient,
     )
+
+The exports below are resolved lazily on first attribute access, so importing
+this package does not load the sandbox worker runtime (and its Azure Identity
+and gRPC dependencies) unless those APIs are actually used.
 """
 
-from durabletask.azuremanaged.preview.sandboxes.client import SandboxActivitiesClient
-from durabletask.azuremanaged.preview.sandboxes.helpers import SandboxActivity
-from durabletask.azuremanaged.preview.sandboxes.worker_profiles import SandboxWorkerProfile
-from durabletask.azuremanaged.preview.sandboxes.worker_profiles import SandboxWorkerProfileOptions
-from durabletask.azuremanaged.preview.sandboxes.worker_profiles import sandbox_worker_profile
-from durabletask.azuremanaged.preview.sandboxes.worker import SandboxWorker
+from importlib import import_module
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from durabletask.azuremanaged.preview.sandboxes.client import SandboxActivitiesClient
+    from durabletask.azuremanaged.preview.sandboxes.helpers import SandboxActivity
+    from durabletask.azuremanaged.preview.sandboxes.worker_profiles import SandboxWorkerProfile
+    from durabletask.azuremanaged.preview.sandboxes.worker_profiles import SandboxWorkerProfileOptions
+    from durabletask.azuremanaged.preview.sandboxes.worker_profiles import sandbox_worker_profile
+    from durabletask.azuremanaged.preview.sandboxes.worker import SandboxWorker
 
 __all__ = [
     "SandboxWorker",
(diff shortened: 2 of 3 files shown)


andystaples (changes requested):
Needs discussion, see my review on #209

andystaples (approved):

PR #218, opened 2026-07-28
Add durable entity unit testing support

### LINKED ISSUE #184: Enhancement: entity unit-testing support in azure-functions-durable v2.x
## Summary

`azure-functions-durable` v2.x supports host-free unit testing for **orchestrators** and **client (trigger) functions**, matching the patterns in [Unit testing Durable Functions and Durable Task SDKs](https://learn.microsoft.com/en-us/azure/durable-task/durable-functions/durable-functions-unit-testing?tabs=python&pivots=durable-functions). There is **no equivalent story for entities**. This issue tracks adding first-class entity unit-testing support, including the v2.x-new class-based `DurableEntity`.

## Current state (what works today)

- **Orchestrators** — `azure.durable_functions.testing.orchestrator_generator_wrapper` drives an orchestrator generator with mocked task results, and `orchestration_trigger`-decorated functions expose `.orchestrator_function` via `.build().get_user_function()`.
- **Client functions** — `durable_client_input`-decorated functions expose `.client_function`, so the raw trigger function can be invoked directly with a mocked client.

Both of the above are being added in the current v2.0.0b1 work as they are required for v1.x compatibility.

## The gap

There is no documented or supported way to unit test an entity without a running Functions host or a Durable Task backend. Specifically:

1. **No test harness to drive an entity operation.** Given an operation name, an input, and an initial state, there is no simple, supported way to invoke the entity and assert on:
   - the operation's return value,
   - the resulting state after the operation,
   - any signals / orchestrations the operation scheduled.
2. **`entity_trigger` handle exposes `.entity_function`**, mirroring `.orchestrator_function`, but there is no `wrapper`/harness equivalent to `orchestrator_generator_wrapper` that simulates the entity batch-execution + state-shim behavior for a single operation in-process.

## v2.x-specific consideration: class-based entities

`DurableEntity` (class-based entities) are **new in v2.x** and change the authoring shape (operations are methods; state is accessed via the entity context / state shim rather than the v1 function-style `context.get_state()/set_state()`). Any entity testing story needs to cover **both**:

- v1-style function entities (`def entity(context)` / `def entity(ctx, input)`), and
- v2.x class-based `DurableEntity` subclasses.

Design questions to resolve:

- How does a customer instantiate/drive a `DurableEntity` subclass in a unit test — do they construct it directly and call the operation method, and if so how is the entity context (state get/set, `signal_entity`, `start_new_orchestration`, etc.) mocked or provided?
- Should we ship a small `azure.durable_functions.testing` helper (e.g. an entity test harness that wraps the state shim and returns `(result, new_state, actions)`), analogous to `orchestrator_generator_wrapper`, so both entity styles are testable with a consistent API?
- What is the minimal, documented pattern we want customers to use, and should it be added to the Learn article's Python / Durable Functions tab (which currently has no entity testing example)?

## Proposed direction

- Provide an entity unit-testing helper in `azure.durable_functions.testing` that executes a single entity operation in-process (reusing the existing entity state shim) and surfaces the return value, the mutated state, and scheduled actions (signals/orchestrations) for assertions.
- Ensure it works for both function-style and class-based (`DurableEntity`) entities.
- Add unit tests covering both styles and, once the API settles, propose an entity example for the docs.

## Notes

Deferred out of the large v2.0.0b1 PR to keep that change scoped to the orchestrator/client compatibility work. This is a net-new v2.x capability rather than a v1 compatibility requirement.

andystaples:
## Summary
- add `execute_entity()` for host-free testing of v1-style, durabletask-native, and class-based entities
- return operation results, resulting state, and stable typed signal/orchestration actions
- document the testing pattern and cover decorated entity handles, identity, state, actions, deletion, and failures

## Validation
- `python -m pytest tests/azure-functions-durable -m "not dts and not azurite and not functions_e2e" --quiet`
- `python -m flake8 tests/azure-functions-durable`
- `python -m pyright ...` for changed Python files
- `python -m pymarkdown -c .pymarkdown.json scan azure-functions-durable/README.md`

Closes #184

### DIFF
--- a/azure-functions-durable/README.md
+++ b/azure-functions-durable/README.md
@@ -61,8 +61,26 @@ assert outcome.actions == ()
 For an `entity_trigger`-decorated function, pass the exposed entity function:
 
 ```python
+import azure.durable_functions as df
+from azure.durable_functions.testing import execute_entity
+
+
+app = df.DFApp()
+
+
+@app.entity_trigger(context_name="context")
+def counter(context: df.DurableEntityContext) -> None:
+    value = context.get_state(initializer=lambda: 0)
+    value += context.get_input()
+    context.set_state(value)
+    context.set_result(value)
+
+
 entity_function = counter.build().get_user_function().entity_function
 outcome = execute_entity(entity_function, "add", input=2, state=3)
+
+assert outcome.result == 5
+assert outcome.state == 5
 ```
 
 The returned `EntityTestResult` includes the operation result, resulting state,
--- a/azure-functions-durable/README.md
+++ b/azure-functions-durable/README.md
@@ -52,8 +52,8 @@ class Counter(DurableEntity):
 
 outcome = execute_entity(Counter, "add", input=2, state=3)
 
-assert outcome.result == 5
-assert outcome.state == 5
+assert outcome.get_result() == 5
+assert outcome.get_state() == 5
 assert outcome.actions == ()
 ```
 
@@ -78,12 +78,17 @@ def counter(context: df.DurableEntityContext) -> None:
 entity_function = counter.build().get_user_function().entity_function
 outcome = execute_entity(entity_function, "add", input=2, state=3)
 
-assert outcome.result == 5
-assert outcome.state == 5
+assert outcome.get_result() == 5
+assert outcome.get_state() == 5
 ```
 
-The returned `EntityTestResult` includes the operation result, resulting state,
-and typed signal or orchestration-start actions scheduled by the operation.
+The returned `EntityTestResult` provides `get_result()` and `get_state()`
+methods plus typed signal or orchestration-start actions scheduled by the
+operation. Pass `expected_type` when reconstructing a custom payload:
+
+```python
+assert outcome.get_state(expected_type=CounterState) == CounterState(value=5)
+```
 
 ## Links
 
--- a/azure-functions-durable/README.md
+++ b/azure-functions-durable/README.md
@@ -33,6 +33,41 @@ Key capabilities include durable orchestrations and sub-orchestrations, durable
 timers, external events, durable entities, retries, versioning, durable HTTP
 calls (`context.call_http(...)`), recurring scheduled tasks, and history export.
 
+## Unit testing entities
+
+Use `execute_entity()` to run one entity operation in-process without a
+Functions host or Durable Task backend. It supports v1-style entity functions,
+durabletask-native entity functions, and `DurableEntity` subclasses:
+
+```python
+from azure.durable_functions.testing import execute_entity
+from durabletask.entities import DurableEntity
+
+
+class Counter(DurableEntity):
+    def add(self, amount: int) -> int:
+        value = self.get_state(int, 0) + amount
+        self.set_state(value)
+        return value
+
+
+outcome = execute_entity(Counter, "add", input=2, state=3)
+
+assert outcome.result == 5
+assert outcome.state == 5
+assert outcome.actions == ()
+```
+
+For an `entity_trigger`-decorated function, pass the exposed entity function:
+
+```python
+entity_function = counter.build().get_user_function().entity_function
+outcome = execute_entity(entity_function, "add", input=2, state=3)
+```
+
+The returned `EntityTestResult` includes the operation result, resulting state,
+and typed signal or orchestration-start actions scheduled by the operation.
+
 ## Links
 
 - [Changelog](CHANGELOG.md)
--- a/azure-functions-durable/azure/durable_functions/testing/entity.py
+++ b/azure-functions-durable/azure/durable_functions/testing/entity.py
@@ -6,9 +6,9 @@
 from __future__ import annotations
 
 import inspect
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime, timezone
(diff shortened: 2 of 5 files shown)


azure-functions-durable/README.md
```diff
@@ -33,6 +33,41 @@ Key capabilities include durable orchestrations and sub-orchestrations, durable
...
+assert outcome.state == 5
+assert outcome.actions == ()
+```
+
+For an `entity_trigger`-decorated function, pass the exposed entity function:
+
+```python
+entity_function = counter.build().get_user_function().entity_function
+outcome = execute_entity(entity_function, "add", input=2, state=3)
+```
```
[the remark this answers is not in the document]
andystaples: Good catch. I made the example self-contained by defining the DFApp and decorated counter entity, and added assertions showing the resulting operation output and state.

azure-functions-durable/azure/durable_functions/testing/entity.py
```diff
@@ -0,0 +1,208 @@
...
+        if encoded_input is not None
+        else None
+    )
+    operation_input = converter.deserialize(encoded_input, input_type)
+
+    try:
+        result = _invoke_entity(
+            entity_callable, operation, context, operation_input)
+        encoded_result = converter.serialize(result)
+        result = converter.deserialize(encoded_result)
```
berndverst: `execute_entity()` fails for custom payloads when `AZURE_FUNCTIONS_DURABLE_STRICT_TYPING=1`: this untyped `deserialize()` raises `TypeError` for a custom `to_json`/`from_json` result, and the same pattern later affects custom state and action inputs. This makes the helper unusable for supported strict-typing entities. Keep the live result/state/action values while serializing only for validation (or pass concrete expected types), and add strict-mode coverage for all three surfaces.

andystaples: Good catch — reproduced on all three surfaces. Fixed in 5130407 by capturing live result, state, and action values through the test context while retaining the runtime serialization calls for validation. Added strict-mode coverage for a custom typed operation input plus custom result, state, signal input, and orchestration input.

berndverst: The strict-mode `TypeError` is resolved, but recording live references still diverges from runtime semantics. `StateShim.set_state()` and the action methods serialize eagerly, so if entity code mutates an object after `set_state()`/`signal_entity()`/`schedule_new_orchestration()`, the worker commits/sends the pre-mutation snapshots while `execute_entity()` now reports the mutated objects (I reproduced helper values `10/20/30` versus worker values `1/2/3`). Capture a serialized snapshot plus its concrete runtime type at each call, then deserialize with that type for the test result, and add a post-call mutation regression test.

andystaples: Fixed in c82f23b. The helper now reconstructs result, state, and action assertions from the exact payloads already serialized by the runtime paths, using each value concrete runtime type. This preserves strict typing without retaining mutable aliases. Added a regression where entity code mutates objects after set_state, signal, and start: the returned result is 10, while persisted and sent snapshots remain 1, 2, and 3, matching worker semantics.

azure-functions-durable/azure/durable_functions/testing/entity.py
```diff
@@ -0,0 +1,286 @@
...
+def _restore_snapshot(
+        encoded_value: str | None,
+        value: Any,
+        converter: DataConverter,
+) -> Any:
+    if value is None:
+        return None
+        value_type: type[Any] = type(value)
+        return converter.deserialize(encoded_value, value_type)
+
```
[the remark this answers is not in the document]
andystaples: Fixed in e731739. The non-None restoration path is now correctly outside the guard, and the snapshot regression suite passes for results, state, signals, and orchestration inputs, including strict mode and post-call mutation cases.

azure-functions-durable/azure/durable_functions/testing/entity.py
berndverst: Direct custom-object snapshots are now correct, but `_restore_snapshot()` still rejects JSON-serializable values whose wire type differs from their Python type. Under strict mode, a tuple result/state/signal/start input serializes successfully in the production worker (all become JSON arrays), while `execute_entity()` raises `TypeError` on all four because it asks `df_loads` to validate the array as `tuple`. For non-custom JSON payloads, deserialize without an expected type so the helper returns the persisted wire shape; reserve the concrete expected type for custom-object envelopes, and add strict tuple coverage.

andystaples: Thanks — I stepped back from type inference and redesigned this in a727c82. EntityTestResult now stores the exact serialized result and persisted state, with explicit get_result(expected_type=...) and get_state(expected_type=...) accessors. Actions likewise retain serialized inputs and expose get_input(expected_type=...). Untyped tuple access therefore returns the persisted JSON list, while strict custom payloads are reconstructed only when the test supplies their expected type. This removes _restore_snapshot and the testing-specific state/context subclasses entirely.

berndverst (approved):
Verified the serialized-outcome redesign at a727c820. Strict-mode tuple and custom result, state, signal, and orchestration payloads now match production wire snapshots, including post-call mutations.

PR #220, opened 2026-07-28
Add propagate-only distributed tracing for Durable Functions

### LINKED ISSUE #179: Distributed tracing for azure-functions-durable (propagate-only mode in durabletask + host correlation)
## Summary

Wire up distributed tracing for `azure-functions-durable` (2.x) so orchestrator user-code executed in the Python worker correlates to the Durable Functions host's trace, **without** duplicating the spans the host already emits. This requires coordinated changes in **both** `durabletask` (core) and `azure-functions-durable`.

Deferred out of the `2.0.0b1` preview (noted as a known limitation in the package CHANGELOG). Surfaced during PR #155.

## Background (how trace context flows today)

Findings from the WebJobs extension (`azure-functions-durable-extension`) and `durabletask-python`:

- **Inbound context already arrives.** The host delivers the parent trace context via the standard durabletask protobuf field `ExecutionStarted.ParentTraceContext` (W3C `traceparent`/`tracestate`) — `ProtobufUtils.cs`. The Python worker's `_execute_orchestrator` already reads `executionStarted.parentTraceContext`, and `DurableFunctionsWorker.execute_orchestration_request` passes the full request through unchanged. So no inbound wiring is required.
- **The host owns span emission.** The extension emits its own OTel spans for the orchestration/activity/entity lifecycle via `TraceHelper` + a dedicated `ActivitySource` (`StartActivityForNewOrchestration`, etc.), and it **ignores** the trace context the worker puts on `ScheduleTask`/`CreateSubOrchestration` response actions (`ProtobufUtils.cs` only maps `Id/Input/Name/Version/Tags`).
- **durabletask-python would duplicate.** When `opentelemetry` is importable, durabletask-python **unconditionally** emits its own orchestration span (`emit_orchestration_span`), with **no toggle** to disable it. `opentelemetry` is not currently a dependency of `azure-functions-durable`, so the loop is dormant.

Net: naively adding `opentelemetry` to `azure-functions-durable` and letting durabletask's loop run under the WebJobs host would produce **duplicate orchestration spans** (host + worker).

## Goal

- Orchestrator user-code spans created in the Python worker nest under the host's orchestration span (correlation).
- The worker does **not** emit orchestration/activity/entity spans that duplicate the host's.

### `durabletask` (core)

- Add a worker-level "propagate-only" tracing mode: attach the incoming `parentTraceContext` as the active OpenTelemetry context around orchestrator execution, but **suppress** durabletask's own span emission (`emit_orchestration_span` and the deferred client/activity spans).
- Suggested surface: a worker/`ConcurrencyOptions` flag (e.g. `distributed_tracing="propagate-only"` or `emit_trace_spans=False`). Default preserves today's behavior (emit spans) for the DTS backend.
- Unit tests for both modes (emit vs. propagate-only).

### `azure-functions-durable`

- Depend on `durabletask[opentelemetry]` (or document the extra) so OTel is available.
- Configure `DurableFunctionsWorker` to use the core "propagate-only" mode, since the WebJobs host emits the lifecycle spans.
- Verify correlation end-to-end via the E2E harness (an app with OTel configured; assert user-code spans parent to the host orchestration span and that no duplicate orchestration spans appear).

## Open verification items

- Confirm the host's `TraceHelper` span emission runs on the **gRPC out-of-proc** dispatch path (vs. only the classic in-proc backend).
- Confirm whether the `azure-functions-python-worker` already attaches per-invocation trace context around orchestrator invocations (may already cover some user-code correlation independent of durabletask).

## Acceptance criteria

- With OTel enabled in a function app, orchestrator user-code spans correlate to the host's orchestration span.
- No duplicate orchestration/activity/entity spans are produced by the Python worker.
- Core `durabletask` gains a documented, tested switch to run in propagate-only mode; `azure-functions-durable` uses it.
- Known-limitation note removed from the `azure-functions-durable` CHANGELOG once shipped.

andystaples:
## Summary

- add `TaskHubGrpcWorker(..., emit_trace_spans=False)` so host integrations can activate inbound W3C trace context without emitting duplicate Durable Task lifecycle spans
- configure `azure-functions-durable` to use propagate-only tracing and depend on `durabletask[opentelemetry]>=1.9.0`
- add unit coverage for context parentage, cleanup, default span emission, and Functions worker configuration
- add a real Azure Functions host E2E test backed by an OpenTelemetry Collector

## Trace validation

The collector-backed test starts a Durable Functions orchestration through the real Functions host and verifies that:

- the Python user-code span has an exported host span as its parent in the same trace
- the parent is not emitted by the Python `durabletask` instrumentation scope
- no Python `durabletask` orchestration, activity, or entity lifecycle span is emitted for the instance

This preserves the existing `durabletask`/DTS trace shape by default while using the host-owned Durable Functions trace shape in `azure-functions-durable`.

Closes #179

### DIFF
--- a/azure-functions-durable/azure/durable_functions/worker.py
+++ b/azure-functions-durable/azure/durable_functions/worker.py
@@ -40,7 +40,10 @@ class DurableFunctionsWorker(TaskHubGrpcWorker):
         # The Functions converter routes payload serialization through the
         # azure-functions codec (df_dumps/df_loads) so user types round-trip in
         # the wire format the Durable Functions host extension expects.
-        super().__init__(data_converter=DEFAULT_FUNCTIONS_DATA_CONVERTER)
+        super().__init__(
+            data_converter=DEFAULT_FUNCTIONS_DATA_CONVERTER,
+            emit_trace_spans=False,
+        )
 
     def add_named_orchestrator(self, name: str, func: task.Orchestrator[Any, Any]) -> None:
         self._registry.add_named_orchestrator(name, func)
--- a/tests/azure-functions-durable/e2e/test_distributed_tracing_e2e.py
+++ b/tests/azure-functions-durable/e2e/test_distributed_tracing_e2e.py
@@ -49,7 +49,7 @@ def test_user_span_correlates_to_host_without_worker_lifecycle_duplicates(
                 break
         time.sleep(0.5)
 
-    assert len(user_spans) == 1
+    assert len(user_spans) == 1, collector.get_logs()
     user_span = user_spans[0]
     summary = [
         (
--- a/tests/azure-functions-durable/e2e/test_distributed_tracing_e2e.py
+++ b/tests/azure-functions-durable/e2e/test_distributed_tracing_e2e.py
@@ -66,8 +66,7 @@ def test_user_span_correlates_to_host_without_worker_lifecycle_duplicates(
 
     python_lifecycle_spans = [
         span for span in spans
+        if span.get("traceId") == user_span.get("traceId")
         if span.get("scopeName") == "durabletask"
-        and _attribute_value(
-            span, "durabletask.task.instance_id") == instance_id
     ]
     assert python_lifecycle_spans == []
--- a/tests/azure-functions-durable/e2e/test_distributed_tracing_e2e.py
+++ b/tests/azure-functions-durable/e2e/test_distributed_tracing_e2e.py
@@ -1,0 +1,73 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""End-to-end distributed tracing validation using an OTEL collector."""
+
+import time
+from typing import Any
+
+import pytest
+
+from ._harness import FunctionApp, OtelCollector
+
+pytestmark = pytest.mark.functions_e2e
+
+
+def _attribute_value(span: dict[str, Any], key: str) -> str | None:
+    for attribute in span.get("attributes", []):
+        if attribute.get("key") == key:
+            return attribute.get("value", {}).get("stringValue")
+    return None
+
+
+def test_user_span_correlates_to_host_without_worker_lifecycle_duplicates(
+        tracing_app: tuple[FunctionApp, OtelCollector]) -> None:
+    app, collector = tracing_app
+    instance_id = app.start_orchestration("correlated_orchestrator")
+    status = app.wait_for_completion(instance_id)
+    assert status["runtimeStatus"] == "COMPLETED"
+
+    deadline = time.time() + 30
+    spans: list[dict[str, Any]] = []
+    user_spans: list[dict[str, Any]] = []
+    parent_spans: list[dict[str, Any]] = []
+    while time.time() < deadline:
+        spans = collector.get_spans()
+        user_spans = [
+            span for span in spans
+            if span.get("name") == "user-orchestrator"
+            and _attribute_value(span, "test.instance_id") == instance_id
+        ]
+        if user_spans:
+            user_span = user_spans[0]
+            parent_spans = [
+                span for span in spans
+                if span.get("traceId") == user_span.get("traceId")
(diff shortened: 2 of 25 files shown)


azure-functions-durable/azure/durable_functions/worker.py
```diff
@@ -40,7 +40,10 @@ def __init__(self) -> None:
         # The Functions converter routes payload serialization through the
         # azure-functions codec (df_dumps/df_loads) so user types round-trip in
         # the wire format the Durable Functions host extension expects.
-        super().__init__(data_converter=DEFAULT_FUNCTIONS_DATA_CONVERTER)
+        super().__init__(
+            data_converter=DEFAULT_FUNCTIONS_DATA_CONVERTER,
+            emit_trace_spans=False,
```
berndverst, line 46: `emit_trace_spans=False` suppresses only worker spans. Because this PR also makes `durabletask[opentelemetry]` mandatory, `DurableFunctionsClient` now always has OTel available; its inherited `schedule_new_orchestration` and `raise_orchestration_event` paths still create `durabletask` lifecycle spans and replace `parentTraceContext` with the new Python span. In a Functions app using host-owned V2 tracing, starts and events therefore remain duplicated and are re-parented under the SDK span.

Please add equivalent propagate-only control to `TaskHubGrpcClient`/`AsyncTaskHubGrpcClient` (defaulting to span emission for compatibility) and disable it in the Durable Functions clients, or suppress emission around their calls while continuing to inject `get_current_trace_context()`.

tests/azure-functions-durable/e2e/test_distributed_tracing_e2e.py
```diff
@@ -0,0 +1,73 @@
...
+    assert len(parent_spans) == 1, summary
+    assert parent_spans[0].get("scopeName") != "durabletask"
+
+    python_lifecycle_spans = [
+        span for span in spans
+        if span.get("scopeName") == "durabletask"
+        and _attribute_value(
+            span, "durabletask.task.instance_id") == instance_id
+    ]
+    assert python_lifecycle_spans == []
```
berndverst, line 81: This assertion cannot detect client lifecycle spans: it also requires `durabletask.task.instance_id == instance_id`, but the create span generates its own UUID independently from the request instance ID, while raise-event spans use `durabletask.event.target_instance_id`. Existing or regressed client-side emissions therefore pass this test.

Please filter by the captured trace instead: after finding `user_span`, assert that no span with the same `traceId` has `scopeName == "durabletask"`.

azure-functions-durable/azure/durable_functions/worker.py
andystaples: Added `emit_trace_spans` (defaulting to `True`) to both core clients and configured both Durable Functions client variants with `False`. Start requests continue to inject the ambient host context in propagate-only mode, while create and raise-event lifecycle spans are suppressed; sync and async behavior are covered by unit tests.

tests/azure-functions-durable/e2e/test_distributed_tracing_e2e.py
andystaples: Updated the E2E assertion to reject every span on the captured user trace whose instrumentation scope is `durabletask`, without relying on instance-ID attributes. This now detects client create spans as well as worker lifecycle spans.

berndverst (dismissed):
Re-reviewed through `335adcc`. Both prior findings are resolved: Durable Functions clients now suppress SDK lifecycle spans while preserving ambient trace-context injection, and the E2E guard rejects all `durabletask` spans on the captured trace. The subsequent scheduled-entity timing change introduces no actionable issue.

berndverst (approved):
Re-reviewed through `0e260d5`, including the Node 22/Azurite workflow updates and the merge with lazy async-client channel creation. Both prior tracing findings remain resolved, merge conflict resolution preserves propagate-only behavior and lazy initialization, and all current checks—including strict Azure Functions E2E—are green.

PR #155, opened 2026-06-23
Add azure-functions-durable v2: a durabletask-based rewrite of Durable Functions for Python

andystaples:
## Summary

Introduces **`azure-functions-durable` 2.x** as a new sub-package in this repository — a ground-up rewrite of the Azure Durable Functions Python SDK, built on top of the [`durabletask`](https://pypi.org/project/durabletask/) SDK. Existing v1 apps keep working through a compatibility layer, while new apps can adopt durabletask-native APIs directly.

## Why build on durabletask

The 1.x SDK implemented the Durable Functions out-of-process model directly, with its own orchestration action/state model and a JSON protocol tailored to the classic Durable Functions host extension. 2.x instead builds on the `durabletask` runtime — the same gRPC-based core that powers the Durable Task Scheduler (DTS) and the modern Durable Functions host. That means:

- one shared orchestration/entity execution core, serialization pipeline, and retry/versioning implementation instead of a Functions-specific reimplementation;
- Functions users can adopt durabletask-native APIs and patterns directly; and
- less protocol drift between the Python worker and the durable backend.

## What's included

- **Decorator programming model** via `DFApp` / `Blueprint` (the Python v2 model).
- **durabletask-native authoring**: two-argument orchestrators and entities (`def orchestrator(ctx, input)`) and class-based entities (`DurableEntity`), alongside supported **v1-style single-argument** functions.
- **v1 compatibility layer**: `DurableOrchestrationContext` / `DurableEntityContext`, deprecated `DurableOrchestrationClient` method aliases (emitting `DeprecationWarning`), return-type wrappers (`DurableOrchestrationStatus`, `PurgeHistoryResult`, `EntityStateResponse`), `HttpManagementPayload`, `EntityId`, `RetryOptions`, and the `call_http` durable HTTP API.
- **New capabilities** inherited from durabletask: `rewind_orchestration`, plus opt-in `DFApp.configure_scheduled_tasks()` and `DFApp.configure_history_export()`.
- **Packaging & CI**: the `azure-functions-durable` package, unit and end-to-end test suites, and strict pyright type-checking for the new package.

## Programming models supported

| Style | Orchestrator | Entity |
| --- | --- | --- |
| durabletask-native | `def o(ctx, input)` | `def e(ctx, input)` / `DurableEntity` class |
| v1-compatible | `def o(context)` | `def e(context)` |

## Breaking changes vs. `azure-functions-durable` 1.x

- **Python 3.13+ is required** (1.x supported 3.10+).
[...]

PR #221, opened 2026-07-28
Harden Durable HTTP trust boundaries

andystaples:
## Summary

- strip managed identity tokens and sensitive headers when Durable HTTP redirects or polling cross origins
- reject non-HTTP(S) redirect and polling targets
- prevent clients from directly starting the internal HTTP polling orchestrator
- add regression coverage for same-origin, cross-origin, multi-hop, and top-level invocation behavior

## Validation

- `python -m pytest tests\azure-functions-durable -m "not dts and not azurite and not functions_e2e" -q`
- `python -m nox -s typecheck_functions`

Fixes the security issues identified in the Durable HTTP implementation introduced by #155.

### DIFF
--- a/azure-functions-durable/CHANGELOG.md
+++ b/azure-functions-durable/CHANGELOG.md
@@ -14,6 +14,13 @@ the synchronous client into synchronous functions and the asynchronous client
 into coroutine functions. Both clients support scheduled-task and history-export
 APIs without an async-to-sync bridge.
 
+FIXED
+
+- Prevented Durable HTTP calls from forwarding managed identity tokens,
+authorization headers, cookies, or function keys to cross-origin redirect and
+polling targets. Direct client invocation of the internal HTTP polling
+orchestrator is now rejected.
+
 ## 2.0.0b1
 
 First preview (beta) release of `azure-functions-durable` 2.x — a ground-up
--- a/azure-functions-durable/CHANGELOG.md
+++ b/azure-functions-durable/CHANGELOG.md
@@ -17,9 +17,11 @@ APIs without an async-to-sync bridge.
 FIXED
 
 - Prevented Durable HTTP calls from forwarding managed identity tokens,
-authorization headers, cookies, or function keys to cross-origin redirect and
-polling targets. Direct client invocation of the internal HTTP polling
-orchestrator is now rejected.
+authorization headers, cookies, proxy credentials, or function keys to
+cross-origin redirect and polling targets. Function keys are now removed from
+every `202 Accepted` poll, including same-origin polls, because the initial
+function-level key may not authorize the status endpoint. Direct client
+invocation of the internal HTTP polling orchestrator is now rejected.
 
 ## 2.0.0b1
 
--- a/azure-functions-durable/azure/durable_functions/http/builtin.py
+++ b/azure-functions-durable/azure/durable_functions/http/builtin.py
@@ -50,6 +50,7 @@ _HTTP_SCHEMES = frozenset(("http", "https"))
 _CROSS_ORIGIN_SENSITIVE_HEADERS = frozenset((
     "authorization",
     "cookie",
+    "proxy-authorization",
     "x-functions-key",
 ))
 _POLL_SENSITIVE_HEADERS = frozenset(("x-functions-key",))
@@ -130,13 +131,17 @@ class _SecureRedirectHandler(urllib.request.HTTPRedirectHandler):
             msg: str,
             headers: Any,
             newurl: str) -> Optional[urllib.request.Request]:
-        if _http_origin(newurl) is None:
-            raise urllib.error.URLError(
-                f"Refusing redirect to non-HTTP(S) URL {newurl!r}.")
-
         redirected = super().redirect_request(
             req, fp, code, msg, headers, newurl)
-        if redirected is None or _is_same_origin(req.full_url, newurl):
+        if redirected is None:
+            return None
+
+        redirected_url = redirected.full_url
+        if _http_origin(redirected_url) is None:
+            raise urllib.error.URLError(
+                f"Refusing redirect to non-HTTP(S) URL {redirected_url!r}.")
+
+        if _is_same_origin(req.full_url, redirected_url):
             return redirected
 
         for header_map in (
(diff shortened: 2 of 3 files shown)


tests/azure-functions-durable/test_http_builtin_compat.py
```diff
@@ -231,7 +297,7 @@ def test_poll_orchestrator_polls_until_complete():
     assert poll_name == BUILTIN_HTTP_ACTIVITY_NAME
     assert poll_input == {
         "method": "GET",
-        "uri": "http://poll",
+        "uri": "http://x/poll",
         "headers": {"h": "v"},
         "tokenSource": {"resource": "r"},
     }
```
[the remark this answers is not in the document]
andystaples: This expectation is intentional and protects the upstream-compatible behavior that removes `x-functions-key` from every 202 poll, including same-origin polls. The initial endpoint's function-level key may not be valid for the status endpoint, which can require a different key.

azure-functions-durable/azure/durable_functions/http/builtin.py
```diff
@@ -221,20 +300,29 @@ def builtin_http_poll_orchestrator(context: Any) -> Generator[Any, Any, DurableH
...
+        same_origin = _is_same_origin(current_uri, location)
         poll_request: dict[str, Any] = {"method": "GET", "uri": location}
-        # Preserve auth for the polling requests.
         if request.get("headers") is not None:
-            poll_request["headers"] = request["headers"]
-        if request.get("tokenSource") is not None:
+            excluded_headers = _POLL_SENSITIVE_HEADERS
+            if not same_origin:
+                excluded_headers = _CROSS_ORIGIN_SENSITIVE_HEADERS
+            poll_request["headers"] = _without_headers(
```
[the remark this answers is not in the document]
andystaples: I'm keeping this behavior because the Durable Functions host intentionally removes `x-functions-key` from every 202 poll: the initial function-level key often does not authorize the status endpoint, which may require a master key. This matches the upstream `CreateLocationPollRequest` policy and avoids forwarding a route-scoped secret beyond the initial request.

azure-functions-durable/azure/durable_functions/http/builtin.py
```diff
@@ -46,6 +46,14 @@
 # usable ``Retry-After`` header.
 _DEFAULT_POLL_INTERVAL_SECONDS = 1
 
+_HTTP_SCHEMES = frozenset(("http", "https"))
+_CROSS_ORIGIN_SENSITIVE_HEADERS = frozenset((
+    "authorization",
+    "cookie",
+    "x-functions-key",
+))
```
berndverst, line 55: `Proxy-Authorization` is also a standard credential header, but this denylist leaves it intact. A caller-supplied value is copied to cross-origin HTTP redirects and 202 polling targets, exposing the proxy secret; I reproduced it through the real redirect chain. Add `proxy-authorization` here and cover both redirect and polling paths. (`ProxyHandler`-managed credentials are re-added by proxy handling rather than copied as an origin header.)

azure-functions-durable/azure/durable_functions/http/builtin.py
```diff
@@ -79,6 +87,72 @@ def _acquire_bearer_token(resource: str) -> str:
...
+
+    def redirect_request(
+            self,
+            req: urllib.request.Request,
+            fp: Any,
+            code: int,
+            msg: str,
+            headers: Any,
+            newurl: str) -> Optional[urllib.request.Request]:
+        if _http_origin(newurl) is None:
```
berndverst: Validate the target only after `super().redirect_request()` determines that this status/method is actually redirectable. For example, PUT + 302 (or POST + 307) normally raises `HTTPError` for the original 3xx, which this activity returns as a response; an FTP `Location` now raises `URLError` first and fails the activity even though urllib would never follow it. Call `super()` first, return/propagate its result, then validate `redirected.full_url` when a redirect request exists.

azure-functions-durable/CHANGELOG.md
```diff
@@ -14,6 +14,13 @@ the synchronous client into synchronous functions and the asynchronous client
 into coroutine functions. Both clients support scheduled-task and history-export
 APIs without an async-to-sync bridge.
 
+FIXED
+
+- Prevented Durable HTTP calls from forwarding managed identity tokens,
```
berndverst, line 19: This entry omits another user-visible behavior change: `x-functions-key` is now stripped from every 202 poll, including same-origin polls. That can turn a previously authorized status poll into a 401; the author confirmed this is intentional for upstream parity. Please document the all-polls behavior and rationale here so users can assess the compatibility impact.

azure-functions-durable/azure/durable_functions/http/builtin.py
andystaples: Added `proxy-authorization` to the cross-origin sensitive-header set and extended both redirect and 202 polling tests to verify it is removed. Proxy-handler-managed credentials remain unaffected because this only filters headers copied from the request.

azure-functions-durable/azure/durable_functions/http/builtin.py
andystaples: Moved target validation after `super().redirect_request()` and now validate the resulting request URL only when urllib elects to follow the redirect. I also added a PUT + 302 regression test confirming the original `HTTPError` behavior is preserved.

azure-functions-durable/CHANGELOG.md
andystaples: Updated the changelog to state that `x-functions-key` is removed from every 202 poll, including same-origin polls, and explain that the initial function-level key may not authorize the status endpoint.

tests/azure-functions-durable/test_http_builtin_compat.py
```diff
@@ -172,11 +175,82 @@ def _ok(req):
...
+
+def test_non_redirectable_response_preserves_http_error():
+    request = urllib.request.Request(
+        "https://example.com/start", method="PUT")
+
+    with pytest.raises(urllib.error.HTTPError) as raised:
+        _SecureRedirectHandler().redirect_request(
+            request, None, 302, "Found", {}, "ftp://example.com/next")
+
+    assert raised.value.code == 302
```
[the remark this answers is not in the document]
andystaples: The base handler does raise `HTTPError` directly for PUT + 302. In Python 3.13, `HTTPRedirectHandler.redirect_request()` documents and implements that behavior with `raise HTTPError(req.full_url, code, msg, headers, fp)` for unsupported status/method combinations; I also verified this regression passes on both Python 3.13 and 3.14. Keeping the assertion ensures our validation does not replace the original 3xx response with `URLError`.

berndverst (approved):
Re-reviewed the follow-up commit and full diff. The Proxy-Authorization leak and redirect-validation regression are fixed with coverage, and the intentional same-origin x-functions-key polling behavior is now clearly documented. Targeted tests and cross-version redirect probes pass.

PR #226, opened 2026-07-29
Defer async client channel creation until first use

andystaples:
## Summary

Fixes a Python 3.13 failure where constructing `AsyncTaskHubGrpcClient` or `DurableFunctionsClient` could raise:

```
RuntimeError: There is no current event loop in thread 'MainThread'.
```

This surfaced in the Azure Functions CI for #220 after an async test ran between two synchronous durable-client construction tests.

## Root cause

`AsyncTaskHubGrpcClient` eagerly created its `grpc.aio` channel in `__init__`. On Python 3.13, `pytest-asyncio` clears the current event loop after an async test. The next synchronous client construction therefore asked gRPC to bind a channel when no current loop existed.

The same sequence is possible outside tests when an async client is constructed after `asyncio.run()` or otherwise outside the event loop that will perform its RPCs. Creating a replacement default loop would only hide the exception and could bind the channel to the wrong loop.

## Fix

- Defer SDK-owned `grpc.aio` channel and stub creation until the first RPC.
- Create them synchronously within that async operation, binding the channel to the loop that actually uses it.
- Keep caller-provided channels eager and unchanged.
- Make closing a never-used client safe and preserve channel recreation behavior.
- Add regression coverage for construction with no current event loop followed by use under `asyncio.run()`.
- Document the user-visible fix in the core and Azure Functions changelogs.

## Validation

- Core unit suite: 1027 passed, 1 skipped.
- Azure Functions unit suite: 211 passed.
- Core/Azure Managed and Azure Functions pyright checks passed.
- Flake8 passed for the changed core source and tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,9 @@ import paths, `__all__`, `dir()`, and star-imports behave exactly as before.
 
 FIXED
 
+- Fixed `AsyncTaskHubGrpcClient` failing during construction when no current
+event loop was set. SDK-owned async gRPC channels are now created on first use,
+binding them to the event loop that performs the RPC.
 - Fixed the worker allocating one `asyncio` task per queued work item before applying the concurrency limit, which made memory use and event-loop scheduling overhead grow with the queue backlog during bursts. In-flight work item tasks are now bounded by the configured `ConcurrencyOptions` limits.
 - Fixed input/output type discovery raising `TypeError: unhashable type` for handlers that are unhashable callables. A callable object registered through `add_named_activity()`, `add_named_orchestrator()`, or `add_named_entity()` is unhashable whenever its class defines `__eq__` without `__hash__` — most commonly a `@dataclass` with a `__call__` method, since dataclasses default to `eq=True`. Annotations on such handlers are now discovered normally instead of failing.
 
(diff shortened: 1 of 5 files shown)


berndverst (commented):
One low-severity documentation requirement remains; the lazy async-channel implementation itself looks sound.

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@ import paths, `__all__`, `dir()`, and star-imports behave exactly as before.
 
 FIXED
 
+- Fixed `AsyncTaskHubGrpcClient` failing during construction when no current
```
berndverst, line 36: **Low severity — document the inherited Azure Managed fix.** `AsyncDurableTaskSchedulerClient` subclasses this client without overriding channel initialization, so constructing it with no current event loop now also changes from raising to succeeding. That user-visible behavior is currently absent from `durabletask-azuremanaged/CHANGELOG.md`, even though the repository requires cross-package changes to be recorded under every applicable `Unreleased` section. Please add a `FIXED` entry there describing the async scheduler client behavior.

andystaples: Added the inherited `AsyncDurableTaskSchedulerClient` behavior under the Azure Managed `Unreleased` changelog in 55b92b8.

berndverst (approved):
The Azure Managed changelog requirement is addressed. I rechecked the follow-up commit, resolved thread, effective full diff, and unchanged lazy-channel implementation; no actionable issues remain.

PR #228, opened 2026-07-29
Fix async export deletion test flake

andystaples:
## Summary

- wait for the asynchronously signaled export entity deletion before asserting the job is gone
- use bounded polling instead of an immediate read or fixed sleep

## Validation

- `python -m pytest tests/durabletask/extensions/history_export/test_client.py -q`
- repeated `test_async_client_create_list_wait_and_delete` 10 times
- `python -m flake8 tests/durabletask/extensions/history_export/test_client.py`

### DIFF
--- a/tests/durabletask/extensions/history_export/test_client.py
+++ b/tests/durabletask/extensions/history_export/test_client.py
@@ -10,6 +10,7 @@ unique job ID to avoid cross-test interference.
 
 from __future__ import annotations
 
+import asyncio
 import gzip
 import json
 import threading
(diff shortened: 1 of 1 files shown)


berndverst (approved):
Reviewed the async deletion race at 075808d0b75b37d4692494ec64297c45891d1622. `SignalEntity` acknowledges enqueueing rather than completed entity processing, so the prior immediate read was racy. The replacement polls the same observable deletion condition with a monotonic 5-second bound and 50 ms async sleeps, and still fails if deletion never occurs. No production/API compatibility change. Focused test passed repeatedly, including a delayed-deletion probe, and the changed test passes flake8.

berndverst (approved):
Reviewed the async deletion race at 075808d0b75b37d4692494ec64297c45891d1622. SignalEntity acknowledges enqueueing rather than completed entity processing, so the prior immediate read was racy. The replacement polls the same observable deletion condition with a monotonic 5-second bound and 50 ms async sleeps, and still fails if deletion never occurs. No production/API compatibility change. Focused test passed repeatedly, including a delayed-deletion probe, and the changed test passes flake8.

PR #236, opened 2026-07-29
Preserve payloads in v1 status query shims

### LINKED ISSUE #231: Preserve output and custom status in v1 status query shims
## Summary

Fix payload-fetch semantics in the v1-compatible `get_status()`, `get_status_all()`, and `get_status_by()` methods.

## Current behavior

`get_status(show_input=False)` maps `show_input` directly to durabletask's `fetch_payloads` flag. The underlying protocol flag is `getInputsAndOutputs`, so `False` removes not only input but also output, custom status, and failure details.

The collection shims call `get_all_orchestration_states()` with the default `OrchestrationQuery(fetch_inputs_and_outputs=False)`, which can similarly omit payload data expected from the v1 status response.

This means the default v1-compatible calls can return completed or failed statuses without their output or custom status.

## Expected behavior

Compatibility shims should fetch the payload data required to reproduce the v1 response, then apply v1-specific projection locally:

- `show_input=False` suppresses only `input`;
- output remains available for completed instances;
- custom status remains available;
- failure details remain available and continue to populate the compatibility output where applicable;
- `get_status_all()` and `get_status_by()` return their expected payload fields.

Native durabletask APIs and their payload-fetch defaults should remain unchanged.

## Acceptance criteria

- Default `get_status()` retains output, custom status, and failure details while omitting input.
- `show_input=True` includes input as well.
- `get_status_all()` and `get_status_by()` retain v1-compatible payload fields.
- Missing instances remain falsy as today.
- Tests cover running, completed, and failed orchestration states without changing native client defaults.

andystaples:
## Summary

- fetch payloads in the deprecated v1 `get_status()`, `get_status_all()`, and `get_status_by()` shims
- apply `show_input` locally so disabling input does not discard output, custom status, or failure details
- preserve native durabletask payload-fetch defaults and add running, completed, failed, and collection regression coverage

## Testing

- `python -m pytest tests\azure-functions-durable -m "not dts and not azurite and not functions_e2e" --quiet`
- `python -m pyright -p azure-functions-durable\pyrightconfig.json`

Fixes #231

### DIFF
--- a/azure-functions-durable/azure/durable_functions/client.py
+++ b/azure-functions-durable/azure/durable_functions/client.py
@@ -265,16 +265,19 @@ class DurableFunctionsClient(AsyncTaskHubGrpcClient):
         exist, a falsy status is returned rather than ``None``.
 
         The ``show_history`` and ``show_history_output`` flags have no
-        equivalent in durabletask and are ignored; ``show_input`` maps to
-        ``fetch_payloads``.
+        equivalent in durabletask and are ignored. Payloads are fetched to
+        preserve the v1 output and custom-status fields; ``show_input`` controls
+        whether the compatibility wrapper exposes the input.
         """
-        state = await self.get_orchestration_state(instance_id, fetch_payloads=show_input)
-        return DurableOrchestrationStatus.from_orchestration_state(state)
+        state = await self.get_orchestration_state(instance_id, fetch_payloads=True)
+        return DurableOrchestrationStatus.from_orchestration_state(
+            state, include_input=show_input)
 
     @deprecated("get_status_all is deprecated; use get_all_orchestration_states instead.")
     async def get_status_all(self) -> list[DurableOrchestrationStatus]:
         """Deprecated alias for :meth:`get_all_orchestration_states`."""
-        states = await self.get_all_orchestration_states()
+        states = await self.get_all_orchestration_states(
+            OrchestrationQuery(fetch_inputs_and_outputs=True))
         return [DurableOrchestrationStatus.from_orchestration_state(state) for state in states]
 
     @deprecated("raise_event is deprecated; use raise_orchestration_event instead.")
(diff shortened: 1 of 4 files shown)


tests/azure-functions-durable/test_client_compat.py
```diff
@@ -527,21 +533,60 @@ def test_from_durabletask_status_reverse_mapping():
...
+            runtime_status=OrchestrationStatus.FAILED,
+            serialized_output=None,
+            failure_details=SimpleNamespace(message="boom"))
+        with patch.object(client, "get_orchestration_state",
+                          new=AsyncMock(return_value=state)):
+            with pytest.warns(DeprecationWarning):
+                status = await client.get_status("abc")
+        status_json = status.to_json()
+        assert status_json["runtimeStatus"] == "Failed"
+        assert status_json["output"] == "boom"
```
berndverst, line 590: **[P2] Preserve failure text on the public `output` property**

This only verifies `to_json()`. With this exact failed state (`serialized_output=None` plus `failure_details`), `status.output` still returns `None` because that property reads only `serialized_output`; only `to_json()` has the failure-message fallback. Existing v1 callers can consume `status.output` directly, so `get_status()` still drops the failure text on its documented object surface. Please give `output` the same fallback (and preferably have `to_json()` reuse it), then assert `status.output == "boom"` here.

berndverst (changes requested):
One compatibility gap remains: failed statuses expose the failure text only through `to_json()`, not through the documented `.output` property.

azure-functions-durable/azure/durable_functions/client.py
```diff
@@ -265,16 +265,19 @@ async def get_status(
         exist, a falsy status is returned rather than ``None``.
 
         The ``show_history`` and ``show_history_output`` flags have no
-        equivalent in durabletask and are ignored; ``show_input`` maps to
-        ``fetch_payloads``.
+        equivalent in durabletask and are ignored. Payloads are fetched to
+        preserve the v1 output and custom-status fields; ``show_input`` controls
+        whether the compatibility wrapper exposes the input.
```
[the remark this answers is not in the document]
andystaples: Updated the `get_status()` docstring to explicitly state that payload fetching preserves failure details alongside output and custom status.

tests/azure-functions-durable/test_client_compat.py
andystaples: Updated `DurableOrchestrationStatus.output` to fall back to `failure_details.message` and made `to_json()` reuse that property. The failed-state regression now asserts both `status.output` and the serialized output.

berndverst (approved):
Re-reviewed the follow-up at this exact head. The failed-status `output` property now preserves the failure message, `to_json()` reuses the same projection, and both prior threads are resolved. Focused projection probes, unit tests, lint, strict typing, and compatibility checks pass.

PR #237, opened 2026-07-29
Use canonical management URLs for HTTP payloads

### LINKED ISSUE #233: Use canonical management URL construction for HTTP payloads
## Summary

Complete HTTP management payload URL parity instead of reconstructing only a subset of the host's management links.

The binding already provides host-generated management URL metadata, but the current clients ignore `managementUrls` and build URLs locally from a hard-coded route shape.

## Current behavior

- `HttpManagementPayload` omits the host-supported `rewindPostUri`.
- Async and sync clients construct `/runtime/webhooks/durabletask/instances/...` directly instead of using the canonical host-provided URL shape.
- URL construction ignores `Forwarded`, `X-Forwarded-Proto`, and `X-Forwarded-Host`, so responses can expose an internal scheme or authority behind a reverse proxy or ingress.
- Locally reconstructing routes can drift from host changes and configured webhook paths.

## Expected behavior

Centralize management URL construction and preserve all canonical payload fields:

- `statusQueryGetUri`
- `sendEventPostUri`
- `terminatePostUri`
- `rewindPostUri`
- `purgeHistoryDeleteUri`
- `restartPostUri`
- `suspendPostUri`
- `resumePostUri`

Prefer the host-provided management URL metadata where possible, while still replacing the instance placeholder and using the externally visible request origin according to the Durable Functions proxy-header conventions.

## Acceptance criteria

- `rewindPostUri` is present in mapping access, `.urls`, `to_json()`, and serialized check-status responses.
- Async and sync clients generate identical payloads.
- Instance IDs and query parameters remain correctly encoded.
- Request-origin behavior is covered for direct requests, `Forwarded`, and `X-Forwarded-*` headers.
- Existing required query parameters and management URL templates remain intact.
- Tests verify all canonical management URL fields.

andystaples:
## Summary

- build HTTP management payloads from the host-provided URL templates so configured paths and required query parameters remain canonical
- replace the instance placeholder with a fully percent-encoded ID and rewrite only the origin using `Forwarded`, `X-Forwarded-*`, or the direct request URL
- add `rewindPostUri` and share the same construction path between async and sync clients, with per-field fallback for older host payloads

## Design

Host templates remain authoritative for path and query content. When an HTTP request is available, only the scheme and authority are replaced with the externally visible request origin. Missing templates fall back individually to the existing route builder so older extension bundles remain compatible.

## Testing

- `python -m pyright -p azure-functions-durable\pyrightconfig.json`
- `python -m pytest tests\azure-functions-durable -m "not dts and not azurite and not functions_e2e" -q`

Closes #233

### DIFF
--- a/azure-functions-durable/azure/durable_functions/client.py
+++ b/azure-functions-durable/azure/durable_functions/client.py
@@ -25,7 +25,7 @@ from .internal.azurefunctions_grpc_interceptor import (
     AzureFunctionsDefaultClientInterceptorImpl,
 )
 from .internal.serialization import DEFAULT_FUNCTIONS_DATA_CONVERTER
-from .http import HttpManagementPayload
+from .http.http_management_payload import HttpManagementPayload, replace_url_origin
 from .internal.compat.durable_orchestration_status import DurableOrchestrationStatus
 from .internal.compat.entity_state_response import EntityStateResponse
 from .internal.compat.orchestration_runtime_status import OrchestrationRuntimeStatus, to_durabletask_statuses
@@ -40,10 +40,15 @@ def _first_forwarded_value(value: str) -> str:
     return value.split(",", 1)[0].strip().strip('"')
 
 
-def _get_request_origin(request: func.HttpRequest) -> str:
+def _get_request_origin(
+        request: func.HttpRequest,
+        use_forwarded_host: bool) -> str:
     request_url = urlparse(request.url)
     proto = request_url.scheme
     host = request_url.netloc
+    if not use_forwarded_host:
+        return f"{proto}://{host}"
+
     request_headers = cast(Mapping[str, str], request.headers)
     headers = {
         name.lower(): value for name, value in request_headers.items()
@@ -77,18 +82,25 @@ def _build_http_management_payload(
         instance_id: str,
         management_urls: dict[str, str],
         base_url: str,
+        http_base_url: str,
         required_query_string_parameters: str,
+        use_forwarded_host: bool,
         request: func.HttpRequest | None) -> HttpManagementPayload:
     encoded_instance_id = quote(instance_id, safe="")
+    configured_base_url = http_base_url or base_url
     request_origin: str | None = None
     if request is not None:
-        request_origin = _get_request_origin(request)
-        instance_status_url = (
-            f"{request_origin}/runtime/webhooks/durabletask/instances/"
-            f"{encoded_instance_id}")
+        request_origin = _get_request_origin(request, use_forwarded_host)
+        if configured_base_url:
+            management_base_url = replace_url_origin(
+                configured_base_url.rstrip("/"), request_origin)
+        else:
+            management_base_url = (
+                f"{request_origin}/runtime/webhooks/durabletask")
     else:
-        instance_status_url = (
-            f"{base_url.rstrip('/')}/instances/{encoded_instance_id}")
+        management_base_url = configured_base_url.rstrip("/")
+    instance_status_url = (
+        f"{management_base_url}/instances/{encoded_instance_id}")
 
     return HttpManagementPayload(
         instance_id,
@@ -114,6 +126,7 @@ class DurableFunctionsClient(AsyncTaskHubGrpcClient):
     requiredQueryStringParameters: str
     rpcBaseUrl: str
     httpBaseUrl: str
+    useForwardedHost: bool
     maxGrpcMessageSizeInBytes: int
     # The host sends this as a .NET TimeSpan string; it is currently stored
     # as-received (see _parse_client_configuration) and is unused, so the raw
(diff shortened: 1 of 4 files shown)


azure-functions-durable/azure/durable_functions/client.py
```diff
@@ -36,6 +36,68 @@
...
+def _get_request_origin(request: func.HttpRequest) -> str:
+    request_url = urlparse(request.url)
+    proto = request_url.scheme
+    host = request_url.netloc
+    request_headers = cast(Mapping[str, str], request.headers)
+    headers = {
+        name.lower(): value for name, value in request_headers.items()
+    }
+
+    forwarded = headers.get("forwarded")
```
berndverst, line 57: The Durable extension gates forwarded-origin handling behind `extensions.durableTask.httpSettings.useForwardedHost`, whose default is `false`. This helper instead trusts `Forwarded`/`X-Forwarded-*` on every request, and `create_check_status_response()` then copies that origin (plus the management system key) into both the response body and `Location`. For example, an `X-Forwarded-Host: attacker.example` request now produces `Location: https://attacker.example/...?...code=<system-key>` even when the host option is disabled. Please pass an explicit host-provided opt-in/trust flag and gate both header families; otherwise retain the origin from `request.url`. Add a default-off regression test as well.

azure-functions-durable/azure/durable_functions/client.py
```diff
@@ -36,6 +36,68 @@
...
+        management_urls: dict[str, str],
+        base_url: str,
+        required_query_string_parameters: str,
+        request: func.HttpRequest | None) -> HttpManagementPayload:
+    encoded_instance_id = quote(instance_id, safe="")
+    request_origin: str | None = None
+    if request is not None:
+        request_origin = _get_request_origin(request)
+        instance_status_url = (
+            f"{request_origin}/runtime/webhooks/durabletask/instances/"
```
berndverst: The middleware-passthrough/gRPC client binding used by this 2.x package does not serialize `managementUrls`; the current Durable extension `dev` binding sends `httpBaseUrl` instead. Therefore `management_urls` is empty in the supported runtime and this hard-coded fallback is always selected, so the PR still discards configured webhook paths. A host-shaped config with `httpBaseUrl: "http://internal.example/custom/durable"` currently returns `https://public.example/runtime/webhooks/durabletask/instances/...` (and the no-request overload returns `/instances/...`). Please derive fallback paths from `httpBaseUrl`/the canonical base URL, or coordinate a host change that supplies templates and version-gate it, and cover the actual middleware-passthrough payload shape rather than injecting `managementUrls` only in unit tests.

berndverst (changes requested):
Two blocking issues remain: the supported gRPC binding shape bypasses the template path, and forwarded-origin headers are trusted without the host's opt-in gate.

azure-functions-durable/azure/durable_functions/client.py
andystaples: Fixed by adding an optional host-provided `useForwardedHost` flag that defaults to `false`; without opt-in, both forwarded-header families are ignored and the origin comes from `request.url`. I also added a regression test showing a spoofed `X-Forwarded-Host` cannot redirect the check-status payload or `Location`.

azure-functions-durable/azure/durable_functions/client.py
andystaples: Fixed the fallback to prefer the supported binding's `httpBaseUrl` over the legacy `baseUrl`, rewriting only its origin while preserving the configured webhook path. The new host-shaped tests omit `managementUrls` and cover both request and no-request overloads.

azure-functions-durable/azure/durable_functions/client.py
berndverst: This is still incomplete at `e0aa562`. The current Durable extension `dev` head (`91eb6695`) serializes `httpBaseUrl`, max message size, and timeout in the middleware-passthrough binding, but it does not serialize `useForwardedHost`; repo search finds the option only in `HttpOptions`, `HttpApiHandler`, and host tests, with no corresponding host PR. Thus a real binding payload always takes this new default-off branch, so enabling `httpSettings.useForwardedHost` still cannot make Python 2.x honor either forwarded-header family. Please land/reference the host serialization change and test the exact emitted JSON.

Also validate adopted values before building the origin: with a synthetic opt-in payload, `Forwarded: proto=;host=public.example` or `X-Forwarded-Proto: ,https` currently makes the check-status `Location` relative (`/custom/durable/instances/...?...code=...`). Keep the request scheme/host for empty parsed values and have `replace_url_origin()` reject an origin without both scheme and authority.

andystaples: Confirmed both points. Commit `79f1ab5` now preserves request components for empty forwarded values and rejects origins without a scheme and authority; the host serialization and rollout decision is tracked in #244 for @berndverst and @andystaples, so I am leaving this thread open pending that discussion.

andystaples:
The host-side dependency for safely honoring forwarded origins is now tracked upstream: Azure/azure-functions-durable-extension#3503.

Because that extension change may take months to ship and become broadly available, the remaining choices for this PR are:

1. Keep the optional `useForwardedHost` parsing, defaulting to `false`. Current hosts remain secure, but forwarded-origin support stays dormant until the upstream field is deployed.
2. Remove forwarded-origin support from this PR and add it in a later Python release after the upstream contract is broadly available.

The host-independent fixes—canonical `httpBaseUrl` paths, template preservation, instance ID encoding, and `rewindPostUri`—can ship under either choice. @berndverst and I will discuss which rollout is preferable.

berndverst (approved):

PR #238, opened 2026-07-29
Stabilize Azure Functions Durable E2E tests

andystaples:
## Summary

- write history-export artifacts to the system temporary directory instead of beneath the Function app, preventing the Functions file watcher from restarting the host mid-suite
- temporarily skip the six E2E checks that directly depend on Azurite returning messages after an initial visibility delay: V1 and durabletask timers, delayed entity signals, and scheduled tasks
- centralize the skip explanation and link it to Azure/Azurite#2343 so the quarantine is easy to remove when emulator behavior is reliable

The AzureStorage polling interval does not address this failure: the failed runs polled after the messages were due, but Azurite returned empty queue responses. Retry-policy E2Es remain enabled because they have not exhibited the failure and use substantially shorter delays.

## Validation

- `python -m flake8` on the changed E2E files
- focused history-export Functions E2E: 2 passed
- full Functions E2E: 74 passed, 7 skipped (six quarantined tests plus the pre-existing Docker-dependent tracing skip)
- confirmed the Functions host log contains no configuration-change restart during history export

### DIFF
--- a/tests/azure-functions-durable/e2e/_markers.py
+++ b/tests/azure-functions-durable/e2e/_markers.py
@@ -1,0 +1,14 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Shared markers for the Azure Functions Durable E2E suite."""
+
+import pytest
+
+
+azurite_delayed_visibility = pytest.mark.skip(
+    reason=(
+        "Azurite intermittently fails to return queue messages after an initial "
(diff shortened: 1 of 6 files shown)


berndverst (approved):
Reviewed head 6d50cb03ea8e19b8fc9ae6816a8f893979ae03d4 against merge base 50f62e2d565b34069a45257343c158916aee84e6. I found no actionable issues in the complete diff or affected Azure Functions E2E lifecycle and test infrastructure.

PR #245, opened 2026-07-29
Implement v1 show_history status compatibility

### LINKED ISSUE #230: Implement v1 show_history status query compatibility
## Summary

Implement the accepted `show_history` and `show_history_output` options on the v1-compatible `get_status()` API.

## Current behavior

`DurableFunctionsClient.get_status()` accepts both options but explicitly ignores them, and `DurableOrchestrationStatus.history` always returns `None`.

The core client now exposes `get_orchestration_history()`, so the underlying history is available even though the compatibility projection is missing. The limitation is mentioned in `CHANGELOG.md`, but no repository issue tracks completing it.

## Expected behavior

When `show_history=True`, retrieve orchestration history and project it into the v1 `historyEvents` response shape. `show_history_output` should control whether output-bearing fields are included in those history events, matching the Durable Functions HTTP management API behavior.

When `show_history=False`, avoid the additional history request and preserve current lightweight status-query behavior.

## Acceptance criteria

- `get_status(show_history=True)` populates `DurableOrchestrationStatus.history` and serialized `historyEvents`.
- `show_history_output=False` omits event output payloads.
- `show_history_output=True` includes event output payloads.
- `show_history=False` does not request history.
- Missing instances and continued-as-new histories are handled consistently with the v1 API.
- Async tests cover history inclusion, output suppression, and the no-history fast path.
- Remove the corresponding known-limitation entry when support is complete.

andystaples:
Fixes #230

## Summary
- fetch orchestration history only when `get_status(show_history=True)` requests it
- project Durable Task history into the compacted v1 `historyEvents` shape
- honor `show_input` and `show_history_output`, including current continued-as-new history
- preserve the no-history fast path for missing, pending, and default status queries
- document that the shared gRPC history protocol exposes structured `FailureDetails`, but not the independent v1-only `Reason` and `Details` fields

## Testing
- `python -m pytest tests/azure-functions-durable -m "not dts and not azurite and not functions_e2e" -q`
- `python -m flake8 tests/azure-functions-durable`
- `python -m pyright azure-functions-durable/azure/durable_functions/internal/compat/durable_orchestration_status.py azure-functions-durable/azure/durable_functions/internal/compat/history_projection.py`
- `python -m pymarkdown -c .pymarkdown.json scan azure-functions-durable/MIGRATION_GUIDE.md`

### DIFF
--- a/azure-functions-durable/azure/durable_functions/internal/compat/history_projection.py
+++ b/azure-functions-durable/azure/durable_functions/internal/compat/history_projection.py
@@ -173,8 +173,6 @@ def _serialize(value: Any) -> Any:
 def _field_name(name: str) -> str:
     if name == "scheduled_start_timestamp":
         return "ScheduledStartTime"
-    if name in {"orchestration_span_id", "span_id"}:
-        return "".join(part.title() for part in name.split("_"))[:-2] + "ID"
     return "".join(part.title() for part in name.split("_"))
 
 
--- a/azure-functions-durable/azure/durable_functions/internal/compat/history_projection.py
+++ b/azure-functions-durable/azure/durable_functions/internal/compat/history_projection.py
@@ -1,0 +1,191 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import json
+from dataclasses import fields, is_dataclass
+from datetime import datetime, timezone
+from typing import Any, cast
+
+from durabletask import history, task
+from durabletask.client import OrchestrationStatus
+
+from .orchestration_runtime_status import from_durabletask_status
+
+
+def project_history(
+        events: list[history.HistoryEvent],
+        *,
+        show_input: bool,
+        show_history_output: bool) -> list[dict[str, Any]]:
+    """Project Durable Task history events into the v1 status-query shape."""
+    projected: list[dict[str, Any]] = []
+    scheduled_tasks: dict[int, tuple[int, history.TaskScheduledEvent]] = {}
+    scheduled_sub_orchestrations: dict[
+        int, tuple[int, history.SubOrchestrationInstanceCreatedEvent]] = {}
+    removed_indexes: set[int] = set()
+
+    for event in events:
+        if isinstance(
+                event,
+                (history.OrchestratorStartedEvent,
+                 history.OrchestratorCompletedEvent)):
+            continue
+
+        item = _project_event(
+            event,
+            show_input=show_input,
+            show_history_output=show_history_output)
+        projected_index = len(projected)
+
+        if isinstance(event, history.TaskScheduledEvent):
+            scheduled_tasks[event.event_id] = (projected_index, event)
+        elif isinstance(
+                event,
+                (history.TaskCompletedEvent, history.TaskFailedEvent)):
+            _add_scheduled_event_data(
+                item,
+                event.task_scheduled_id,
+                scheduled_tasks,
+                removed_indexes,
+                show_input)
+        elif isinstance(event, history.SubOrchestrationInstanceCreatedEvent):
+            scheduled_sub_orchestrations[event.event_id] = (
+                projected_index, event)
+        elif isinstance(
+                event,
+                (history.SubOrchestrationInstanceCompletedEvent,
+                 history.SubOrchestrationInstanceFailedEvent)):
+            _add_scheduled_event_data(
+                item,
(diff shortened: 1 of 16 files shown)


azure-functions-durable/azure/durable_functions/internal/compat/history_projection.py
```diff
@@ -0,0 +1,191 @@
...
+        return {key: _serialize(item) for key, item in mapping.items()}
+    return value
+
+
+def _field_name(name: str) -> str:
+    if name == "scheduled_start_timestamp":
+        return "ScheduledStartTime"
+    if name in {"orchestration_span_id", "span_id"}:
+        return "".join(part.title() for part in name.split("_"))[:-2] + "ID"
+    return "".join(part.title() for part in name.split("_"))
```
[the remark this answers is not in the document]
andystaples: Updated `_field_name()` to emit the v1-compatible `SpanId` and `OrchestrationSpanId` casing, with regression coverage for both projected trace fields.

azure-functions-durable/azure/durable_functions/internal/compat/history_projection.py
```diff
@@ -0,0 +1,189 @@
...
+    item["FunctionName"] = scheduled_event.name
+    if show_input and scheduled_event.input is not None:
+        item["Input"] = scheduled_event.input
+    removed_indexes.add(scheduled_index)
+
+
+def _serialize(value: Any) -> Any:
+    if isinstance(value, datetime):
+        return _format_datetime(value)
+    if isinstance(value, task.FailureDetails):
```
berndverst, line 151: **Compatibility:** For `TaskFailedEvent` and `SubOrchestrationInstanceFailedEvent`, this emits only nested `FailureDetails`, while the v1 history contract exposes top-level `Reason` and `Details`. A migrated status viewer reading those established keys will fail exactly when an activity or child orchestration fails. Preserve the legacy fields through the history protocol/model (without bypassing the host's details-redaction behavior), emit them here alongside structured details, and add failed-event parity tests.

berndverst (changes requested):
One blocking v1 compatibility issue remains in failed activity and sub-orchestration history projection.

azure-functions-durable/azure/durable_functions/internal/compat/history_projection.py
andystaples: Preserved the host-provided failure aliases in the typed history model: `TaskFailureDetails.errorMessage` now becomes top-level `Reason`, and its optional host-redacted `stackTrace` becomes `Details`, while `FailureDetails` remains available. Added protocol-deserialization tests for task and sub-orchestration failures plus v1 projection parity coverage.

berndverst: Thanks for adding coverage, but this still cannot preserve the v1 fields at `387cd1e`. DurableTask.Core persists `Reason`, `Details`, and `FailureDetails` independently, while the host's `ToHistoryEventProto` sends only `FailureDetails` for both failed event types. `_failure_event_kwargs()` therefore loses both aliases when structured details are absent, and when they are present it substitutes `FailureDetails.StackTrace` for the separately gated legacy `Details` value. On the exact head, I reproduced both cases for task and sub-orchestration failures: no `failureDetails` projects only `EventType`/`Timestamp`, while a structured stack trace is promoted verbatim to top-level `Details`.

Please carry optional `reason`/`details` through the protocol and host mapper (populating `details` only according to the host's existing `IncludeDetails` decision), retain `FailureDetails` unchanged, and consume those exact fields here. The tests should cover absent structured details and a redacted/omitted legacy `Details` alongside a populated structured stack trace.

andystaples: I traced the loss to the shared gRPC history protocol: Core stores `Reason`, `Details`, and `FailureDetails` independently, so deriving the legacy fields from structured error message/stack data was inaccurate and could manufacture `Details` when it was intentionally absent. I removed that synthesis; this PR now preserves canonical `FailureDetails` exactly, tests both present and absent structured details, and documents the v1-only field gap as a known compatibility limitation without requesting upstream protocol changes.

berndverst (approved):
The unsafe failure-field synthesis has been removed, the shared gRPC protocol boundary is now explicitly documented and tested, and no actionable issues remain at this head.

Issue #222, opened 2026-07-28
Add task hub and instance ID prefix orchestration query filters


andystaples:
## Summary

The protobuf `InstanceQuery` supports `taskHubNames` and `instanceIdPrefix`, and the .NET Durable Task client exposes both filters. Python's `OrchestrationQuery` and request builder currently omit them.

## Current behavior

Python orchestration queries support creation-time bounds, runtime status, result limits, and payload inclusion. There is no way to:

- restrict results to one or more task hubs;
- restrict results to instance IDs beginning with a prefix.

The generated protobuf already contains both fields, so no protocol update is required.

## Expected parity

Extend `OrchestrationQuery` with Python-style equivalents of:

- `task_hub_names: list[str] | None`
- `instance_id_prefix: str | None`

Serialize these through `build_query_instances_req()` to `InstanceQuery.taskHubNames` and `InstanceQuery.instanceIdPrefix` for both sync and async clients.

.NET references:

- [`OrchestrationQuery`](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Client/Core/OrchestrationQuery.cs)
- [`GrpcDurableTaskClient.GetAllInstancesAsync`](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Client/Grpc/GrpcDurableTaskClient.cs)

## Acceptance criteria

- Callers can filter orchestration queries by task hub names.
- Callers can filter orchestration queries by instance ID prefix.
- Both fields are serialized into the existing protobuf request.
- Existing queries remain backward compatible when the fields are omitted.
- Unit tests cover each filter independently and together.

PR #251, opened 2026-08-10
Add instance ID prefix orchestration queries

andystaples:
## Summary
- add `instance_id_prefix` to `OrchestrationQuery` and serialize it to `InstanceQuery.instanceIdPrefix` 
- cover protobuf serialization and in-memory query filtering
- document the public query filter

## Validation
- `python -m pytest tests/durabletask/test_batch_actions.py` 
- `python -m flake8 tests/durabletask`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+ADDED
+
+- Added `OrchestrationQuery.instance_id_prefix` to retrieve orchestration
+instances whose IDs begin with a specified prefix.
+
 ## v1.9.0
 
 ADDED
--- a/durabletask/client.py
+++ b/durabletask/client.py
@@ -192,11 +192,11 @@ class OrchestrationQuery:
     created_time_from: datetime | None = None
     created_time_to: datetime | None = None
     runtime_status: list[OrchestrationStatus] | None = None
-    instance_id_prefix: str | None = None
     # Some backends don't respond well with max_instance_count = None, so we use the integer limit for non-paginated
     # results instead.
     max_instance_count: int | None = (1 << 31) - 1
     fetch_inputs_and_outputs: bool = False
+    instance_id_prefix: str | None = None
 
 
 @dataclass
--- a/durabletask/client.py
+++ b/durabletask/client.py
@@ -192,6 +192,7 @@ class OrchestrationQuery:
     created_time_from: datetime | None = None
     created_time_to: datetime | None = None
     runtime_status: list[OrchestrationStatus] | None = None
+    instance_id_prefix: str | None = None
     # Some backends don't respond well with max_instance_count = None, so we use the integer limit for non-paginated
     # results instead.
     max_instance_count: int | None = (1 << 31) - 1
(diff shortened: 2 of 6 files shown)


durabletask/client.py
```diff
@@ -192,6 +192,7 @@ class OrchestrationQuery:
     created_time_from: datetime | None = None
     created_time_to: datetime | None = None
     runtime_status: list[OrchestrationStatus] | None = None
+    instance_id_prefix: str | None = None
```
berndverst: This inserts a field into the middle of a public dataclass's generated positional constructor. Existing `OrchestrationQuery(None, None, None, None, True)` calls now bind `True` to `max_instance_count` and leave `fetch_inputs_and_outputs=False`; numeric fourth arguments can reach `StringValue` as the prefix and raise `TypeError`. Please move `instance_id_prefix` after `fetch_inputs_and_outputs` and add a regression test for the original five-position mapping.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+ADDED
+
+- Added `OrchestrationQuery.instance_id_prefix` to retrieve orchestration
+instances whose IDs begin with a specified prefix.
```
berndverst, line 13: This API is also exposed through `DurableTaskSchedulerClient` and the sync/async Durable Functions clients by inheritance. The repository changelog policy therefore requires matching `## Unreleased` entries in `durabletask-azuremanaged/CHANGELOG.md` and `azure-functions-durable/CHANGELOG.md`. Please document the inherited user-facing filter in both provider changelogs.

berndverst (changes requested):
Requesting changes for the positional-constructor compatibility regression. The wire mapping and query behavior otherwise look sound.

durabletask/client.py
andystaples: Moved `instance_id_prefix` after the existing fields and added a regression test that verifies the original five-position constructor mapping.

CHANGELOG.md
andystaples: Added matching `## Unreleased` entries to the Durable Task Scheduler and Azure Functions Durable changelogs.

andystaples:
Resolves one half of #222

CHANGELOG.md
berndverst: Verified at `c71cec2`: both provider changelogs now document the inherited prefix filter under `Unreleased`.

durabletask/client.py
berndverst: Verified at `c71cec2`: `instance_id_prefix` is appended after all existing fields, so the original five positional bindings and existing five-position class patterns remain unchanged. The regression test passes.

berndverst (dismissed):
Second-pass review complete at `c71cec2`. The positional constructor and existing class-pattern order are preserved, both provider changelogs now cover the inherited API, and the existing protobuf field is wired through the shared sync/async request path without client-side filtering. Targeted tests, affected-file Flake8, and strict core/Functions Pyright checks pass.

berndverst (approved):
Third-pass review complete at `b9953dd`. The only delta from approved `c71cec2` is the merge of current `main` (`9a660a1`, suspend/resume reasons); the two feature commits are unchanged. The appended dataclass field still preserves the original five-argument constructor and class-pattern prefix, both provider changelogs remain present, and protobuf field 7 continues through the shared sync/async request builder without client-side filtering. Current-head CI (24 checks), the full batch-action test file (16 tests), affected-file Flake8, strict targeted core/Azure Managed and full Functions Pyright, and runtime constructor/pattern/wire checks pass.

PR #253, opened 2026-08-10
Add custom FailureDetails properties

### LINKED ISSUE #67: Add support for custom properties in FailureDetails
## Context

The original reference is [Azure/azure-functions-durable-extension#3215](https://github.com/Azure/azure-functions-durable-extension/pull/3215), which added exception-property enrichment for .NET isolated Durable Functions on top of [microsoft/durabletask-dotnet#474](https://github.com/microsoft/durabletask-dotnet/pull/474) and the `TaskFailureDetails.properties` protobuf field from [microsoft/durabletask-protobuf#45](https://github.com/microsoft/durabletask-protobuf/pull/45).

## Research update (2026-07-28)

| SDK | Current state |
| --- | --- |
| .NET | Complete. `IExceptionPropertiesProvider.GetExceptionProperties(Exception)` returns `IDictionary<string, object?>?`; the provider is registered on the worker, invoked recursively for the exception chain, and serialized through `map<string, google.protobuf.Value>`. The public `TaskFailureDetails` exposes both `InnerFailure` and `Properties`. [PR #474](https://github.com/microsoft/durabletask-dotnet/pull/474) established the behavior; [PR #560](https://github.com/microsoft/durabletask-dotnet/pull/560) later added a sample without changing the contract. |
| Java | Complete since [microsoft/durabletask-java#263](https://github.com/microsoft/durabletask-java/pull/263) (2026-04-29). Java added `ExceptionPropertiesProvider`, worker-builder registration, recursive `innerFailure`, protobuf round-tripping, integration tests, a depth limit, and safe handling when the provider itself throws. |
| JavaScript | Custom properties are not implemented. The main [`FailureDetails`](https://github.com/microsoft/durabletask-js/blob/19781fd3efeaae0043584b53e9949105cfddad15/packages/durabletask-js/src/task/failure-details.ts) still exposes only message, error type, and stack trace, and no code reads or writes the generated protobuf properties map. JS does have partial adjacent support: [PR #310](https://github.com/microsoft/durabletask-js/pull/310) added recursive `Error.cause` serialization, and the entity-specific failure model reads inner failures recursively. This is context, not a dependency for Python. |
| Python | The generated protobuf gained `properties` incidentally with [PR #110](https://github.com/microsoft/durabletask-python/pull/110), so no protobuf update is needed. Python now writes recursive `innerFailure` ([PR #99](https://github.com/microsoft/durabletask-python/pull/99)), made `FailureDetails` a frozen dataclass for JSON-safe history export ([PR #172](https://github.com/microsoft/durabletask-python/pull/172)), and added fully-qualified error types plus `is_caused_by()` ([PR #194](https://github.com/microsoft/durabletask-python/pull/194)). However, all public proto-to-model paths still discard `innerFailure` and `properties`, and workers cannot populate custom properties. |

The current .NET and Java implementations agree on the portable property model: `None`/null, strings, booleans, numbers, nested string-keyed mappings, and lists are represented with `google.protobuf.Value`; unsupported values fall back to their string representation. This path is independent of the payload `DataConverter`. Protobuf numbers are float64, so integer-like values read back as `float`. Date/time values are not inferred on read: review of extension PR #3215 rejected prefix-based `dt:`/`dto:` type guessing because ordinary customer strings could be irreversibly coerced. Python should require providers to stringify date/time values and treat prefixed values from .NET as opaque strings.

## Target behavior

Add an optional Python `ExceptionPropertiesProvider` protocol with:

```python
def get_exception_properties(
    self, exception: Exception
) -> Mapping[str, Any] | None: ...
```

When configured on a worker, it is called for the top-level exception and each captured inner exception. `None` or an empty mapping omits the wire map. If the provider raises, log a warning with the exception and continue reporting the task failure without custom properties, matching Java's worker-safety behavior.

Extend the existing frozen `FailureDetails` dataclass by appending backward-compatible defaulted fields:

```python
inner_failure: FailureDetails | None = None
properties: dict[str, Any] | None = None
```

Keep stored properties as ordinary JSON-safe dictionaries so `dataclasses.asdict()` and history export continue to work after PR #172.

## Implementation plan

1. **Add shared failure conversion utilities.** In `durabletask/internal/helpers.py`, add Python-to-`google.protobuf.Value` and protobuf-to-Python conversion for null, bool, number (including `Decimal`, encoded as float64), string, nested mapping, and list values, with string fallback for unsupported outbound values. Do not add date/time type inference or prefix parsing. Extend `new_failure_details()` to accept an optional provider/logger, preserve the existing cycle/depth protection, and apply the provider recursively.
2. **Expose the complete public model everywhere.** Add `inner_failure` and `properties` to `durabletask.task.FailureDetails`, then centralize protobuf-to-`FailureDetails` conversion and reuse it from `TaskFailedError`, `client.parse_orchestration_state()`, `history._failure_details()`, and entity failure conversion. Also teach `_failure_details_from_core_dict()` to consume legacy `Properties` and recursive `InnerFailure`. This covers task failures, orchestration state queries, entity failures, and all failure-bearing history events without separate partial implementations.
3. **Configure workers.** Add `exception_properties_provider` to `TaskHubGrpcWorker`, retain it on the worker, and thread it through `_OrchestrationExecutor` / `_RuntimeOrchestrationContext` so user-orchestrator failures are enriched. Use it for activity failures, entity-operation failures, output-serialization failures, and outer worker failure paths that create `TaskFailureDetails`.
4. **Preserve provider support in derived workers.** Add and forward the same optional parameter from `DurableTaskSchedulerWorker`; export the provider protocol from the public package surface. Existing constructors and behavior remain unchanged when no provider is supplied.
5. **Keep Azure Functions integration non-blocking.** The new `azure-functions-durable` 2.x package was added after this issue was opened and creates internal workers per invocation, while activity exceptions are reported by the Functions host/worker pipeline. Core read support will make incoming properties visible immediately. A Functions-specific registration/serialization surface should be designed separately rather than blocking core and DTS support in this issue.
6. **Add coverage.** Test every protobuf value kind, `Decimal`/float64 conversion, and nested combination; null/empty mappings; unsupported-value string fallback; recursive properties on `__cause__`/`__context__`; cycle/depth limits; selective providers; provider exceptions; and no-provider compatibility. Add worker-level tests for orchestration, activity, and entity failures; round-trip tests through `TaskFailedError`, orchestration state, and history events; a history JSON-serialization regression test; an Azure Managed constructor/forwarding test; and a fixture that consumes .NET-produced properties.
7. **Document the user-facing API.** Add a short worker example showing correlation/domain metadata extraction and update `CHANGELOG.md` plus `durabletask-azuremanaged/CHANGELOG.md` under `Unreleased`.

## Acceptance criteria

- A configured provider can attach portable custom values to orchestration, activity, and entity failures.
- Properties and recursive inner failures survive protobuf round-trips and are visible through `TaskFailedError.details`, orchestration-state APIs, and history APIs.
- Existing three-argument `FailureDetails(...)` construction and workers without a provider behave unchanged.
- Provider failures never replace or suppress the original task failure.
- No protobuf regeneration or JavaScript implementation is required before merging Python support.

andystaples:
## Summary
- Add `ExceptionPropertiesProvider` and worker configuration for portable exception metadata.
- Preserve recursive failure details and custom properties across task exceptions, orchestration state, history, and entity failures.
- Document supported values and the intentional `dt:` / `dto:` .NET parity gap; add core and Azure Managed changelog entries.

## Validation
- `python -m pytest tests\durabletask\test_failure_details.py tests\durabletask\entities\test_entity_failure_handling.py`
- `python -m pytest tests\durabletask-azuremanaged\test_azuremanaged_grpc_resiliency.py`
- `python -m flake8 tests\durabletask`
- `python -m flake8 durabletask-azuremanaged`
- `python -m flake8 tests\durabletask-azuremanaged`
- `python -m pyright` on changed production files
- `python -m pymarkdown -c .pymarkdown.json scan README.md`

Resolves #67


durabletask/worker.py
```diff
@@ -60,6 +60,14 @@
 _WorkItem = tuple[Callable[..., Any], Callable[..., Any], tuple[Any, ...], dict[str, Any]]
 
 
+class ExceptionPropertiesProvider(Protocol):
+    """Extract portable custom properties from an exception."""
+
+    def get_exception_properties(self, exception: Exception) -> Mapping[str, Any] | None:
+        """Return properties to include in the exception's failure details."""
+        ...
```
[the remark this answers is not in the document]
andystaples: The ellipsis is intentional: it is the no-op method body required by this structural Protocol declaration.

tests/durabletask/entities/test_entity_failure_handling.py
```diff
@@ -147,6 +148,41 @@ def test_orchestrator(ctx: task.OrchestrationContext, _):
...
 
+def test_entity_failure_properties_are_available_to_orchestrators():
+    class PropertiesProvider:
+        def get_exception_properties(self, exception: Exception):
+            return {"message": str(exception), "retryable": False}
+
+    def failing_entity(ctx: entities.EntityContext, _):
+        raise ValueError("Something went wrong!")
+
+    def test_orchestrator(ctx: task.OrchestrationContext, _):
```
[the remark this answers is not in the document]
andystaples: Added an explicit `return None` after the try/except.

durabletask/internal/helpers.py
```diff
@@ -141,32 +146,125 @@ def get_qualified_name(t: type) -> str:
...
-        innerFailure=new_failure_details(inner, _visited) if inner else None
+        innerFailure=(
+            new_failure_details(
+                inner, exception_properties_provider, logger, _visited)
+            if inner else None)
     )
+    if properties:
+        for key, value in cast(Mapping[Any, Any], properties).items():
+            failure_details.properties[str(key)].CopyFrom(protobuf_value_from_python(value))
+    return failure_details
```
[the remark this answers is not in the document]
andystaples: Fixed in 242afd6: invalid provider results and conversion errors are logged and discarded so the original failure is still reported.

durabletask/worker.py
```diff
@@ -1615,7 +1628,8 @@ def set_failed(self, ex: Exception | pb.TaskFailureDetails):
             self.next_sequence_number(),
             pb.ORCHESTRATION_STATUS_FAILED,
             None,
-            ph.new_failure_details(ex) if isinstance(ex, Exception) else ex,
+            ph.new_failure_details(
```
berndverst: Unhandled task failures lose the metadata this PR just captured. When an activity, entity, or sub-orchestration failure escapes the orchestrator, `ex` is a `TaskFailedError`; rebuilding it here serializes only that wrapper and calls the provider on `TaskFailedError`, while its existing `details` (the original properties and inner chain) are discarded. With a provider that returns `{"code": "VALUE"}` only for `ValueError`, the activity failure has the property, but the final `OrchestrationState.failure_details` has `properties=None` and `inner_failure=None`. Please carry the underlying task failure into the orchestration failure (for example as its inner failure) and add an unhandled-task regression test.

durabletask/task.py
```diff
@@ -481,6 +481,8 @@ class FailureDetails:
     message: str
     error_type: str
     stack_trace: str | None
+    inner_failure: FailureDetails | None = None
+    properties: dict[str, Any] | None = None
```
berndverst, line 488: One core serialization path still drops these fields: `durabletask.extensions.history_export.serialization.orchestration_state_to_dict()` manually emits only message/type/stack and one inner level, so exported state metadata omits `properties` and truncates deeper causes, while failure-bearing history events include them. Please use a shared recursive `FailureDetails` serializer there and cover state metadata with properties plus a multi-level inner chain.

berndverst (changes requested):
Requesting changes because an unhandled failed task drops its original custom properties and inner-failure chain when it fails the parent orchestration. Core history-export state metadata also omits the new fields.

durabletask/worker.py
andystaples: Fixed in ddaa948. Unhandled TaskFailedError instances now retain the original protobuf TaskFailureDetails as the orchestration failure's inner failure, preserving custom properties and the complete existing inner-failure chain. Added a regression test.

durabletask/task.py
andystaples: Fixed in ddaa948. History export now uses a recursive FailureDetails serializer for orchestration-state metadata, including properties and every inner-failure level. Added multi-level coverage.

durabletask/worker.py
berndverst: Verified at `cc0cd89`: `_RuntimeOrchestrationContext.set_failed()` now preserves the original protobuf failure as `innerFailure`, including the complete nested chain and custom properties. The regression test and an additional multi-level propagation check both pass.

durabletask/task.py
berndverst: Verified at `cc0cd89`: orchestration-state history export now uses the recursive serializer, preserving properties and all inner-failure levels. The multi-level export coverage passes.

durabletask/task.py
```diff
@@ -574,16 +576,24 @@ class TaskFailedError(Exception):
...
+                    details.properties[key].CopyFrom(value)
+        self._failure_details = details
+        self._details = pbh.failure_details_from_protobuf(details)
 
     @property
     def details(self) -> FailureDetails:
         return self._details
 
+    @property
+    def failure_details(self) -> pb.TaskFailureDetails:
```
berndverst: **Public API boundary:** This accessor is only needed by _RuntimeOrchestrationContext.set_failed, but its public name and return type expose durabletask.internal.orchestrator_service_pb2.TaskFailureDetails and hand callers the mutable transport object that is later re-emitted. Removing it after release would itself be a compatibility break. Please keep this plumbing private/internal and leave .details as the stable public FailureDetails surface.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ADDED
 
+- Added `ExceptionPropertiesProvider` and the
+  `exception_properties_provider` worker option to attach portable custom
```
berndverst: Non-blocking: repository changelog policy keeps wrapped entry text unindented. Please align these continuation lines with the existing entry below and make the same adjustment in durabletask-azuremanaged/CHANGELOG.md.

berndverst (changes requested):
Second pass at cc0cd89: the two previously reported propagation/export regressions are verified fixed. Requesting changes for the remaining public API-boundary issue; the changelog-format note is non-blocking.

CHANGELOG.md
andystaples: Fixed while resolving the current main merge in 1fcdd38. Both changelog entries now use unindented continuation lines, matching repository policy.

durabletask/task.py
andystaples: Fixed in 4e8e6ff. Removed the public `failure_details` accessor; the protobuf transport detail is now private and used only by worker internals. `TaskFailedError.details` remains the stable public FailureDetails API.

berndverst (approved):

durabletask/worker.py
```diff
@@ -1610,11 +1623,22 @@ def set_failed(self, ex: Exception | pb.TaskFailureDetails):
         # self._pending_actions.clear()  # Cancel any pending actions
         self._completion_status = pb.ORCHESTRATION_STATUS_FAILED
 
+        if isinstance(ex, task.TaskFailedError):
```
berndverst, line 1626: **Post-merge follow-up (distinct from the direct-unhandled fix above):** this special case only preserves transport details when `TaskFailedError` is the root exception. If an orchestrator catches one and does `raise RuntimeError("wrapper") from task_error`, recursive serialization treats the cause as an ordinary Python exception; the final chain is only `RuntimeError -> TaskFailedError`, and the original failure type, properties, and inner chain are lost. Please preserve retained task transport details when a `TaskFailedError` appears anywhere in a cause/context chain and cover the wrapped case.

README.md
```diff
@@ -39,6 +39,34 @@ pip install durabletask[azure-blob-payloads]
...
+    def get_exception_properties(self, exception: Exception):
+        return {"error_code": getattr(exception, "error_code", None)}
+
+
+worker = TaskHubGrpcWorker(exception_properties_provider=FailureProperties())
+```
+
+Properties support `None`, booleans, numbers, strings, nested mappings, and
+lists. They are available from `TaskFailedError.details`, orchestration state,
+and history as `FailureDetails.properties`.
```
berndverst, line 62: **Post-merge follow-up:** this guarantee does not currently hold across failed sub-orchestrations in `InMemoryOrchestrationBackend`. Its watcher rebuilds the parent event from `Exception(errorMessage)`, so a child state that retains `builtins.ValueError`, custom properties, and an inner chain reaches the parent as `builtins.Exception` with `properties=None` and no inner failure. Please copy the child's `TaskFailureDetails` into the failed event and add an in-memory parent/child regression test.

PR #252, opened 2026-08-10
Add caller-provided logger configuration

andystaples:
## Why this is needed

Azure Functions 2.x clients and workers currently receive private core-SDK loggers with SDK-owned INFO stream handlers. Those records bypass the Azure Functions host logging pipeline, so host-level configuration cannot reliably control their level, routing, sampling, or telemetry correlation.

This change makes Functions-created components use the `azure.durable_functions.client` and `azure.durable_functions.worker` module loggers. They have no SDK-installed handlers and propagate normally, so the Functions host owns logging behavior as it does for the v1 SDK.

## Changes

- route Azure Functions client and worker logs through the host-managed module logger hierarchy
- add an optional caller-provided `logger` to core and Azure Managed clients and workers
- deprecate `log_handler` and `log_formatter` while preserving their existing behavior
- document shared and isolated logger patterns, including handler and propagation trade-offs
- add regression coverage for logger forwarding, configuration preservation, legacy warnings, and Functions host logging

## Validation

- `python -m pytest tests\durabletask\test_logging.py tests\durabletask-azuremanaged\test_logging.py`
- `python -m pytest tests\azure-functions-durable\test_logging.py tests\azure-functions-durable\test_client_lifecycle.py tests\azure-functions-durable\test_client_compat.py tests\azure-functions-durable\test_worker_compat.py`
- `python -m flake8 durabletask tests\durabletask`
- `python -m flake8 durabletask-azuremanaged tests\durabletask-azuremanaged`
- `python -m flake8 azure-functions-durable tests\azure-functions-durable`
- `python -m pyright durabletask\client.py durabletask\worker.py durabletask\internal\shared.py durabletask-azuremanaged\durabletask\azuremanaged\client.py durabletask-azuremanaged\durabletask\azuremanaged\worker.py`
- `python -m pymarkdown -c .pymarkdown.json scan docs\features.md`

PR #256, opened 2026-08-10
Close worker background event loops

andystaples:
## Why this is needed

`TaskHubGrpcWorker.start()` creates an asyncio event loop on its background thread but never closes it. Stopped workers therefore retain loop resources until cyclic garbage collection, which can emit a delayed `ResourceWarning: unclosed event loop` during unrelated code.

This surfaced as the intermittent Python 3.11 failure in PR #252 when garbage collection ran inside a warning-capture block.

## Changes

- use `asyncio.run()` to own and close the worker thread's event loop
- add deterministic regression coverage that retains the loop and verifies it is closed after the thread exits
- document the resource leak fix in the core changelog

## Validation

- `python -m pytest tests\durabletask\test_worker_resiliency.py tests\durabletask\test_worker_concurrency_loop.py tests\durabletask\test_worker_concurrency_loop_async.py --quiet`
- `python -m nox -s core_tests-3.11 -- tests\durabletask\test_worker_resiliency.py -k worker_start --quiet`
- `python -m flake8 tests\durabletask`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,9 @@ import paths, `__all__`, `dir()`, and star-imports behave exactly as before.
 
 FIXED
 
+- Fixed `TaskHubGrpcWorker` leaving its background event loop unclosed after
+shutdown, which could retain resources and emit delayed `ResourceWarning`
+messages.
 - Fixed `AsyncTaskHubGrpcClient` failing during construction when no current
 event loop was set. SDK-owned async gRPC channels are now created on first use,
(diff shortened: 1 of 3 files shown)


berndverst (approved):
Reviewed the event-loop lifecycle change, surrounding worker restart/stop behavior, regression coverage, compatibility, performance, changelog, and validation results. The change is correct, non-breaking, and appropriately closes worker-owned event-loop resources.

PR #270, opened 2026-09-24
Add configurable large payload storage to the Azure Functions provider

andystaples:
## Summary

- Add `DFApp.configure_large_payloads(payload_store=...)`, shared across synchronous and asynchronous clients, workers, activities, entities, and imported blueprints.
- Keep external references JSON-compatible with the Functions host and hydrate nested legacy entity results. Resolve worker configuration at invocation time for blueprints imported before configuration.
- Document Blob Storage configuration and retention requirements, and add the provider changelog entry. No version or dependency-minimum changes.
- Add unit and real-host E2E coverage for inline and 300 KB payloads, both client types, activities, replay, custom status, entities, external events, sub-orchestrations, and continue-as-new.

## Validation

- Provider unit suite: 277 passed.
- Non-tracing Functions E2E suite through Nox: 83 passed, 6 skipped, including all 7 new large-payload cases.
- Strict provider Pyright: 0 errors.
- Maintained provider source, tests, and Nox configuration: flake8 passed.
- README Markdown lint and whitespace checks passed.

E2E used Azure Storage with Azurite; DTS-specific behavior and distributed tracing were not exercised. Existing lint errors in local generated build copies were left untouched.


azure-functions-durable/azure/durable_functions/internal/payloads.py
```diff
@@ -0,0 +1,94 @@
...
+        return json.dumps(await self._store.upload_async(data, instance_id=instance_id))
+
+    def _unwrap(self, token: str) -> str:
+        if self._store.is_known_token(token):
+            return token
+        try:
+            value = json.loads(token)
+        except (ValueError, TypeError):
+            return token
+        return value if isinstance(value, str) else token
```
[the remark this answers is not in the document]
andystaples: The ambiguity is real: JSON-quoting references for the Functions host makes an otherwise ordinary JSON string indistinguishable from a reference. One correction: `BlobPayloadStore` validates the `blob:v1:<container>:<blobName>` structure with nonempty components, rather than accepting every string with the prefix.

In a1d8ca0, we have made this an explicit documented restriction of opting into payload externalization, rather than changing the protobuf/host contract or introducing an escaping protocol. Whole payload strings recognized by the configured store are reserved references, including JSON-quoted references, regardless of the size threshold. Download failures propagate rather than falling back to literal data.

Applications that need to carry a reference as literal data can wrap it in an object, e.g. `{"reference": "blob:v1:container:blob"}`, and retain that wrapper across durable payload boundaries. Added tests verify whole-string reference interpretation, raw/quoted missing-reference failures, and preservation of object-wrapped existing and missing references both inline and when the wrapper itself is externalized. The provider Nox unit suite passes: 283 tests.

A more elaborate marker or hash would not distinguish a deliberately round-tripped reference from a transport reference; arbitrary-data support would require escaping or separate metadata. We propose accepting the documented reserved-string contract for this opt-in feature. Leaving this thread open for reviewer agreement; the collision has not been eliminated.

***This comment was written by an agent on behalf of andystaples***

azure-functions-durable/azure/durable_functions/internal/converters.py
```diff
@@ -136,11 +137,12 @@ def decode(cls, data: meta.Datum, *,
         # carrying a custom-object envelope surfaces as TypeError below and is
         # re-raised as ValueError.
         if data_type in ['string', 'json']:
+            value = deexternalize_payload(data.value)
```
berndverst: [P1] Keep Blob I/O off the Functions invocation event loop

This call performs a synchronous download, and `encode()` below performs a synchronous upload. The Functions worker calls binding decode/encode directly inside its async invocation handler, before/after the thread-pool dispatch even when the activity itself is synchronous. A 150 ms fake-store operation delayed an unrelated 10 ms coroutine for the full 150 ms in both directions. Blob latency and retries therefore stall unrelated invocations and completions across the worker process.

Please keep the synchronous converter hooks free of network I/O and move hydration/externalization into invocation wrappers that can await the store's native async methods or offload synchronous work. `asyncio.to_thread()` is available at the provider's Python minimum and propagates `contextvars`, but it must be awaited outside these synchronous hooks; cancelling that await does not stop an already-running thread.

Upstream dispatch contract: [input conversion](https://github.com/Azure/azure-functions-python-worker/blob/27d44b8888b91d38e18f3cd73877b888c708c6d5/runtimes/v2/azure_functions_runtime/handle_event.py#L195-L209), [function dispatch](https://github.com/Azure/azure-functions-python-worker/blob/27d44b8888b91d38e18f3cd73877b888c708c6d5/runtimes/v2/azure_functions_runtime/handle_event.py#L236-L248), and [output conversion](https://github.com/Azure/azure-functions-python-worker/blob/27d44b8888b91d38e18f3cd73877b888c708c6d5/runtimes/v2/azure_functions_runtime/handle_event.py#L272-L279).

azure-functions-durable/azure/durable_functions/internal/serialization.py
```diff
@@ -42,7 +44,7 @@ def serialize(self, value: Any) -> str | None:
     def deserialize(self, data: str | None, target_type: type | None = None) -> Any:
         if data is None or data == "":
             return None
-        return df_loads(data, expected_type=target_type)
+        return df_loads(deexternalize_payload(data), expected_type=target_type)
```
berndverst: [P1] Separate retryable storage failures from application failures

Hydrating legacy entity results inside `deserialize()` puts Blob I/O inside `_OrchestrationExecutor.execute()`'s application-error catch. When a stored entity result raises `ConnectionError`, the worker returns an `OrchestratorResponse` with `ORCHESTRATION_STATUS_FAILED`, although the entity operation succeeded. Recovery of storage cannot resume that terminal orchestration.

Please introduce a host-compatible retry/abandon path for retryable storage failures, distinct from permanent reference errors and user-code exceptions. Moving the read before replay alone is not sufficient: the [current Functions host](https://github.com/Azure/azure-functions-durable-extension/blob/702ec544213eb5b3016237d1390ed4c19b92e8aa/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs#L292-L314) also converts ordinary unclassified invocation exceptions into orchestration failures. This needs to be addressed at the provider/host boundary rather than assuming that raising any exception retries the work item.

azure-functions-durable/azure/durable_functions/client.py
```diff
@@ -178,6 +179,7 @@ def __init__(self, client_as_string: str):
             interceptors=interceptors,
             channel_options=channel_options,
             data_converter=DEFAULT_FUNCTIONS_DATA_CONVERTER,
+            payload_store=get_transport_payload_store(),
```
berndverst: [P2] Hydrate legacy entity envelopes in history retrieval too

Passing this store to the base client only hydrates protobuf `StringValue` fields whose entire content is a reference. Functions entity history contains JSON protocol envelopes, for example an `eventRaised.input` object whose `result` field contains the JSON-encoded reference. Both configured sync and async `get_orchestration_history()` currently return that nested reference unchanged: `collect_history_events()` and its async counterpart never invoke the new `FunctionsDataConverter.deserialize()` hook.

History exports and compatibility history queries therefore retain these Blob references despite using the configured Python client. Please add protocol-aware hydration for entity request/result envelopes in history ingestion, sharing preprocessing with worker execution where appropriate. Identify actual entity protocol events rather than recursively interpreting arbitrary application object strings, which must remain literal under the documented wrapper contract.

For a provider-local implementation, a `TypedDict` for the validated envelope and a reusable `TypeIs` guard (Python 3.13) could make this boundary clearer. The guard must genuinely validate the shape, and event correlation is still needed to distinguish protocol messages from application data.

berndverst (commented):
I recommend addressing the correctness and concurrency findings before merging. The inline comments cover activity I/O blocking the worker event loop, storage failures being classified as application failures, incomplete entity-history hydration, and loss of output-size error details. The override and JSON parsing suggestions are non-blocking.

Additional Python 3.13 guidance:

- Keep newer APIs provider-local: `azure-functions-durable` requires 3.13, but the shared `durabletask` package still supports 3.10.
- If provider-side hydration batches independent downloads, `asyncio.TaskGroup` plus bounded concurrency can provide structured cleanup. It cancels and awaits siblings when a child fails, then propagates failures as an exception group. Python 3.13 improves simultaneous cancellation handling. This belongs around storage I/O, not inside replayed user orchestrators, which must use durable scheduling APIs.
- `asyncio.timeout()` can express an overall async transfer deadline if that is part of the intended configuration contract. Coordinate it with Blob SDK retries; do not add an arbitrary hard-coded deadline. Cancelling an await on `asyncio.to_thread()` does not stop an already-running thread.
- I do not see a concrete need for `copy.replace()`, `typing.ReadOnly`, or generic-syntax churn in this diff. Avoid simply caching `get_transport_payload_store()`: workers and blueprints may be created before configuration, so caching `None` would break the supported initialization order.

I am treating the documented reserved whole-string token syntax as an intentional opt-in restriction, rather than repeating the existing collision finding.

azure-functions-durable/azure/durable_functions/worker.py
```diff
@@ -84,6 +86,7 @@ def _register_entity_once(self, func: task.Entity[Any, Any]) -> None:
             self._registered_entity_functions[name] = func
 
     def execute_orchestration_request(self, func: task.Orchestrator[Any, Any], context: Any) -> str:
+        self._payload_store = get_transport_payload_store()
```
berndverst: [P2] Preserve the cause when output externalization fails

Enabling the core worker's externalization also enables an error path that this host-driven wrapper cannot report accurately. When an orchestration returns a value above `max_stored_payload_bytes`, core `_execute_orchestrator()` catches and logs the size-limit `ValueError` without calling `CompleteOrchestratorTask`. This wrapper then raises `RuntimeError('Orchestrator execution did not produce a response.')`. For example, a 202-byte serialized result with a 150-byte maximum loses the useful limit diagnosis from the reported invocation failure. The entity wrapper has the analogous missing-response path.

Please preserve or explicitly translate the original deterministic error into useful failure details instead of replacing it with the missing-response error. After fixing propagation, `BaseException.add_note()` (available since Python 3.11) can attach the operation/instance context while preserving the original exception type and traceback. Notes alone will not fix the swallowed exception, and should not include payload contents or credentials.

azure-functions-durable/azure/durable_functions/internal/payloads.py
```diff
@@ -0,0 +1,94 @@
...
+        raise ValueError("A different payload store is already configured in this process")
+    _payload_store = payload_store
+
+
+def get_payload_store() -> PayloadStore | None:
+    """Return the app's store, or None when externalization is disabled."""
+    return _payload_store
+
+
+class _FunctionsPayloadStore(PayloadStore):
```
berndverst: [Non-blocking] Mark the provider's overrides explicitly

Since this package requires Python 3.13, please consider `from typing import override` and `@override` on the `PayloadStore` implementations here (`options`, the sync/async upload/download methods, and `is_known_token`). The same applies to `FunctionsDataConverter`'s overrides. This makes the intended base-class contract explicit and lets Pyright catch an override whose base member disappears or is renamed, without adding a runtime dependency. For `options`, use `@property` above `@override`.

Keep this provider-local; importing `typing.override` directly in the shared Python 3.10-compatible core would raise its minimum requirement.

azure-functions-durable/azure/durable_functions/internal/payloads.py
```diff
@@ -0,0 +1,94 @@
...
+        return json.dumps(self._store.upload(data, instance_id=instance_id))
+
+    async def upload_async(self, data: bytes, *, instance_id: str | None = None) -> str:
+        return json.dumps(await self._store.upload_async(data, instance_id=instance_id))
+
+    def _unwrap(self, token: str) -> str:
+        if self._store.is_known_token(token):
+            return token
+        try:
+            value = json.loads(token)
```
berndverst: [Non-blocking] Avoid parsing ordinary large JSON values just to reject them as references

After rejecting a raw token, `_unwrap()` fully parses every JSON value and then discards the result unless it is a string. Because `is_known_token()` is called throughout protobuf traversal, this adds an unnecessary full parse/allocation for ordinary objects and arrays, and again when already-hydrated values reach the Functions codec.

After the raw-token check, consider returning early unless the first non-JSON-whitespace character is a double quote. A JSON-string candidate still needs `json.loads()` so escapes and the existing raw/quoted reference contract remain correct. This preserves application objects while avoiding the redundant parse; no new dependency or newer Python feature is required.

azure-functions-durable/azure/durable_functions/internal/serialization.py
andystaples: Agreed that this is a real failure mode, and that moving hydration before replay is not sufficient: an ordinary Python invocation exception does not tell the Functions host to abandon/redeliver the work item. We're deferring the storage-recovery policy rather than treating the error-propagation fixes as a solution to it.

I compared the standalone implementations and updated my local durabletask-dotnet checkout to current main (`cfe26800bbb693d3d4ddaa74ca7a5c3b457a9dc1`):

- Standalone Python hydrates inputs outside the user-code failure catch; escaping errors reach its work-item manager's explicit abandon path. Completion/output-upload failures are logged without completing the work item. That relies on the standalone transport lifecycle, which the Functions wrapper does not run.
- Standalone .NET performs payload I/O in async gRPC interceptors. Its [sidecar interceptor](https://github.com/microsoft/durabletask-dotnet/blob/cfe26800bbb693d3d4ddaa74ca7a5c3b457a9dc1/src/Extensions/AzureBlobPayloads/Interceptors/AzureBlobPayloadsSideCarInterceptor.cs) distinguishes permanent storage failures (including payload limits, and most HTTP 4xx except 408/429) from other errors. Activity/orchestration permanent output failures become useful failure responses; other errors propagate to the [worker processor](https://github.com/microsoft/durabletask-dotnet/blob/cfe26800bbb693d3d4ddaa74ca7a5c3b457a9dc1/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs), which has explicit abandon RPCs. That is work-item recovery, not an added user activity retry policy.
- The .NET Functions orchestration executor instead uses `GrpcOrchestrationRunner.LoadAndRun` and returns a response to the host. The inspected path does not establish an equivalent SDK-managed payload-recovery guarantee through those standalone completion interceptors. Backend-managed large-message handling is also a separate feature, so I would not claim accepted cross-SDK parity for this limitation.

For this PR, we're keeping the bounded fixes separate: preserve useful storage/size errors, avoid storage I/O in activity binding converters, and hydrate correlated entity history envelopes. Sync activities remain sync; async activities await async storage. No additional durable activity retry policy, new transient-error classifier, timeout, or abandon behavior is being introduced. The documentation explicitly warns that a transient storage error can still become a terminal orchestration failure.

A follow-up needs an agreed host-compatible recovery signal and error classification, coordinated across SDK/host implementations and tested with actual storage failures. Storage transport retries, work-item redelivery, and user-configured activity retries should remain distinct. Leaving this thread open to make that limitation explicit.

***This comment was written by an agent on behalf of andystaples***

azure-functions-durable/azure/durable_functions/internal/payloads.py
andystaples: Applied in 305d50d. After checking for a raw token, _unwrap() returns immediately unless the first non-JSON-whitespace character is a double quote. Objects, arrays, numbers, booleans, and null therefore bypass json.loads() during reference detection. JSON-string candidates still use json.loads() so escaping and the raw/quoted reference contract are preserved.

Hydration continues to restore serialized payload text for the normal codec to decode; the reference check no longer constructs and discards ordinary JSON objects or arrays.

***This comment was written by an agent on behalf of andystaples***

azure-functions-durable/azure/durable_functions/client.py
andystaples: Fixed in 305d50d, with the preprocessing shared with worker execution in 7415068. Both sync and async history retrieval hydrate entity request inputs and correlated result envelopes. The helper validates legacy entity request metadata and correlates replies by request ID, including modern EntityOperationCalled events; it does not recursively interpret arbitrary application object strings.

Regression coverage exercises both client variants, correlated envelopes, malformed/unrelated events, signals, and preservation of object-wrapped literal references. Worker tests additionally cover correlation across past/new history and hydration before replay. The Functions unit suite and large-payload host E2E tests pass.

***This comment was written by an agent on behalf of andystaples***

azure-functions-durable/azure/durable_functions/internal/payloads.py
andystaples: Applied in 305d50d. The provider's `PayloadStore` adapter and `FunctionsDataConverter` implementations now use `typing.override`; `options` has `@property` above `@override`. This remains provider-local, with no `typing.override` import or Python-minimum change in the shared core. The provider's Python 3.13 Pyright check passes with zero errors.

***This comment was written by an agent on behalf of andystaples***

azure-functions-durable/azure/durable_functions/internal/converters.py
andystaples: Fixed in 305d50d and completed in 7415068. Binding converters and the Functions data converter no longer perform storage I/O. Registered orchestration, entity, and activity wrappers await the store's native async download/upload methods, including when the user's function is synchronous. Nested entity payloads are hydrated before replay rather than during deserialization.

Synchronous user code runs on execution threads with contextvars and invocation logging context preserved; the runtime's exported pool is reused when available. Regression tests verify async-only storage, nonblocking awaits, and that a second download can complete while the only execution thread is occupied. Validation: 332 Functions unit tests passed; 83 host E2E tests passed, 7 skipped, including tracing without Docker. Host logging and category-filtering tests passed. This does not claim to solve the separate host retry/abandon concern.

***This comment was written by an agent on behalf of andystaples***

azure-functions-durable/azure/durable_functions/worker.py
andystaples: Fixed in 305d50d and retained in the async entrypoints added by 7415068. The Functions wrapper captures the core completion response first, then externalizes it outside the core execution method's catch. Storage and payload-size failures therefore propagate as the original exceptions instead of being replaced by the missing-response RuntimeError.

Parameterized regression tests cover orchestration/entity responses, synchronous/asynchronous entrypoints, and storage/size-limit failures. They verify the original storage exception identity and preservation of the size-limit diagnosis. This preserves error details without introducing a storage-recovery or retry policy.

***This comment was written by an agent on behalf of andystaples***

